PR #5, opened 2016-07-06
Add js-class-names rule

### LINKED ISSUE #4: Lint static literal usage of js- class names
`js-foo` should be okay. `js-foo-${bar}` should be a lint error.

josh:
Closes #4.

-
CC: @mislav @dgraham

### DIFF
--- a/lib/rules/js-class-name.js
+++ b/lib/rules/js-class-name.js
@@ -1,0 +1,46 @@
+module.exports = function(context) {
+  var allJsClassNameRegexp = /\bjs-[_a-zA-Z0-9-]*/g
+  var validJsClassNameRegexp = /^js(-[a-z0-9]+)+$/g
+  var endWithJsClassNameRegexp = /\bjs-[_a-zA-Z0-9-]*$/g
+
+  function checkStringFormat(node, str) {
+    var matches = str.match(allJsClassNameRegexp) || []
+    matches.forEach(function(match) {
+      if (!match.match(validJsClassNameRegexp)) {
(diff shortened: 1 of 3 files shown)


lib/rules/js-class-name.js
```diff
@@ -0,0 +1,46 @@
...
+  }
+
+  return {
+    "Literal": function(node) {
+      if (typeof node.value === 'string') {
+        checkStringFormat(node, node.value)
+
+        if (node.parent &&
+            node.parent.type === 'BinaryExpression' &&
+            node.parent.operator === '+' &&
```
mislav: What's with the plus operator?

lib/rules/js-class-name.js
```diff
@@ -0,0 +1,46 @@
...
+    var matches = str.match(allJsClassNameRegexp) || []
+    matches.forEach(function(match) {
+      if (!match.match(validJsClassNameRegexp)) {
+        context.report(node, "js- class names should be lowercase and only contain dashes.")
+      }
+    })
+  }
+
+  function checkStringEndsWithJSClassName(node, str) {
+    if (str.match(endWithJsClassNameRegexp)) {
```
mislav: Would `str.endsWith(allJsClassNameRegexp)` work?

lib/rules/js-class-name.js
josh: Looking for `"foo" + bar`

lib/rules/js-class-name.js
josh: I *think* `endsWith` only accepts strings, not a regexp.

PR #8, opened 2016-11-10
Package Requirements

josh:
I want to start to document the requirements around the npm packages we use at GitHub.

[Rendered Markdown](https://github.com/github/eslint-plugin-github/blob/package-requirements/docs/package-requirements.md)

<hr>

To: @github/js 
CC: @github/design-systems

### DIFF
--- a/bin/npm-check-github-package-requirements.js
+++ b/bin/npm-check-github-package-requirements.js
@@ -1,0 +1,42 @@
+#!/usr/bin/env node
+
+const fs = require('fs')
+const path = require('path')
+
+const checks = []
+function check(name, callback) {
+  checks.push([checks.length+1, name, callback])
+}
+
+function run() {
+  process.stdout.write(`1..${checks.length}\n`)
+  checks.forEach(([count, name, callback]) => {
+    Promise.resolve().then(callback).then(() => {
+      process.stdout.write(`ok ${count} - ${name}\n`)
+    }).catch(error => {
+      process.stdout.write(`not ok ${count} - ${name}\n  ${error}\n`)
+    })
+  })
+}
+
+
+const packageRoot = process.argv[2]
+
+check('package.json exists', () => {
+  const packageJsonPath = path.join(packageRoot, 'package.json')
+
+  if (!fs.existsSync(packageJsonPath)) {
+    throw new Error('package.json does not exist')
+  }
+})
+
+check('package.json license is set', () => {
+  const packageJsonPath = path.join(packageRoot, 'package.json')
+  const pkg = JSON.parse(fs.readFileSync(packageJsonPath, 'utf8'))
+
+  if (!pkg.license) {
+    throw new Error('license not set')
+  }
+})
+
+run()
--- a/docs/package-requirements.md
+++ b/docs/package-requirements.md
@@ -1,0 +1,21 @@
+# Package Requirements
+
+A set of criteria used to evaluate packages used by [GitHub](https://github.com).
+
+* MUST be published on [npm registery](https://www.npmjs.com).
+* SHOULD be also available on [GitHub](https://github.com) with tagged releases.
+* MUST provide an [OSI](https://opensource.org/licenses) license file and declare in `package.json` `"license"` field.
+* SHOULD be available under an MIT license.
+* MUST follow [SemVer](http://semver.org) versioning.
+* MUST be installable via [npm](https://www.npmjs.com/package/npm) or [yarn](https://www.npmjs.com/package/yarn) toolchain.
+
+## Browser specific requirements
+
+Additional criteria for packages that are used at runtime rather than at build time.
+
+* MUST provide a `package.json` `"main"` pointing to an ES5 browser compatible source file using the UMD format.
+* SHOULD provide a `package.json` `"jsnext"` point to an ES6 browser compatible source file using ES6 exports.
+* `"main"` and `"jsnext"` source files MUST not `require()` node specific dependencies.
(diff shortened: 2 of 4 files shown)


josh:
This list is focused on third party packages, however, any package we publish will also follow these rules.

I want to follow up this document with a set of recommendations for publishing GitHub open source npm packages. It will be more specific about what technologies to use and how to layout directories, etc.

docs/package-requirements.md
```diff
@@ -0,0 +1,22 @@
+# Package Requirements
+
+A set of criteria used to evaluate packages used by [GitHub](https://github.com).
+
+* MUST be published on [npm registery](https://www.npmjs.com).
+* SHOULD be also available on [GitHub](https://github.com) with tagged releases.
+* MUST provide an [OSI](https://opensource.org/licenses) license file and declare in `package.json` `"license"` field.
+* SHOULD be available under an MIT license.
```
mislav: This is unnecessarily restrictive. I don't see why we would frown upon seeing either CC0 or Unlicense here. Should we list a few licenses that we prefer?

docs/package-requirements.md
```diff
@@ -0,0 +1,22 @@
...
+* MUST follow [SemVer](http://semver.org) versioning.
```
mislav: This is difficult to ascertain. A project may say that they are following SemVer, but we can't know how closely they are sticking to the principle. Or, they may be following SemVer, but not have that explicitly stated anywhere. (This is the case with my own projects.) I guess SemVer states that you should have a "semver" tag in your project, but I rarely see that in practice.

docs/package-requirements.md
```diff
@@ -0,0 +1,22 @@
...
+* MUST be installable via [npm](https://www.npmjs.com/package/npm) or [yarn](https://www.npmjs.com/package/yarn) toolchain.
```
mislav: Isn't this the same as "MUST be published on npm registery"?

docs/package-requirements.md
```diff
@@ -0,0 +1,22 @@
...
+
+## Browser specific requirements
+
+Additional criteria for packages that are used at runtime rather than at build time.
+
+* MUST provide a `package.json` `"main"` pointing to an ES5 browser compatible source file using the UMD format.
```
mislav: Are we also going to look out for `"browser"` property, even though its adoption might be a mess right now?

docs/package-requirements.md
```diff
@@ -0,0 +1,22 @@
...
+* SHOULD provide a `package.json` `"jsnext"` point to an ES6 browser compatible source file using ES6 exports.
+* `"main"` and `"jsnext"` source files MUST not `require()` node specific dependencies.
+* MUST only declares `package.json` `"dependencies"` that are used at runtime in a browser. Nonessential build packages are declared under `"devDependencies"`.
+* MUST not depend on [jQuery](https://jquery.com).
+* MUST not set or leak any global variables.
```
mislav: Unless it's a polyfill.

docs/package-requirements.md
josh: Yeah, maybe I can just remove this line. *Most* packages follow some kind of reasonable versioning policy.

docs/package-requirements.md
josh: Ha, yeah, I was thinking about that too, but I didn't have any spec to reference. I'd like to do more research before including it. But definitely a possibility.

docs/package-requirements.md
josh: Yep, good call. I'll add some notes about polyfills in general.

docs/package-requirements.md
mislav: Well, if you're keeping the clause that it must be published on npm registry, then that implies that the package is also installable via npm. We could maybe change this line to "SHOULD be installable via yarn".

josh:
Any other feedback on this document?

broccolini:
Thanks for the ping. This seems fine to me, @jonrohan what do you think? Wondering if we should link to this in the [contribution docs](https://github.com/github/github/pull/63571)?

bin/npm-check-github-package-requirements.js
```diff
@@ -0,0 +1,42 @@
...
+})
+
+check('package.json license is set', () => {
+  const packageJsonPath = path.join(packageRoot, 'package.json')
+  const pkg = JSON.parse(fs.readFileSync(packageJsonPath, 'utf8'))
+
+  if (!pkg.license) {
+    throw new Error('license not set')
+  }
+})
```
josh: @mislav checked in an initial script to checking a few things about the package. Not really interesting yet. Just prints out a list of checks in the TAPs format.

bin/npm-check-github-package-requirements.js
```diff
@@ -0,0 +1,42 @@
...
+
+const checks = []
+function check(name, callback) {
+  checks.push([checks.length+1, name, callback])
+}
+
+function run() {
+  process.stdout.write(`1..${checks.length}\n`)
+  checks.forEach(([count, name, callback]) => {
+    Promise.resolve().then(callback).then(() => {
```
josh: I was maybe just going to move those to helper functions.

mislav:
> I was hoping this TAP writer would be simple. Do you think its just worth using mocha itself without the tap formatter?

Nah I think Mocha is overkill here. It's simple to generate a TAP writer, but you just have to ensure that no output lines will start un-indented unless they are `ok/not ok` lines.

PR #9, opened 2016-12-12
Lint for JS messing with CSRF tokens

mastahyeti:
To prevent people from shooting themselves in the foot with our per-form CSRF tokens.

/cc @github/appsec @github/js


docs/rules/authenticity_token.md
```diff
@@ -0,0 +1,65 @@
+# `<input name="authenticity_token">`
```
mastahyeti: @josh Feel free to commit to this branch if you'd like to change anything in these docs. After we merge this PR, I'll copy the docs into our [security docs](https://github.com/github/github/blob/master/docs/security.md) as well.

tests/authenticity-token.js
```diff
@@ -0,0 +1,37 @@
+var rule = require('../lib/rules/authenticity-token')
+var RuleTester = require('eslint').RuleTester
+
+var ruleTester = new RuleTester()
+
+ruleTester.run("authenticity-token", rule, {
+  valid: [],
+  invalid: [
+    {
+      code: "document.querySelector('form').elements['authenticity_token'].value",
```
mislav: I have a feeling that `elements.authenticity_token` syntax would still be allowed, since we're only checking string literals?

mastahyeti: Yeah, I think that would be allowed. Any idea what type of AST node that is?

PR #13, opened 2017-04-18
Add a rule for no-dataset usage

nakajima:


lib/rules/no-dataset.js
```diff
@@ -0,0 +1,11 @@
+module.exports = function(context) {
+  return {
+    "Identifier": function(node) {
```
josh: I think we want to look for `MemberExpression`s here. `Identifier` is too generic. For an example, this would error if you called a variable `dataset`.

Theres a nice way to find what nodes you want to look for at http://eslint.org/parser/. Basically type the source of bad code and look at the AST for the closest `type`.

lib/rules/no-dataset.js
```diff
@@ -0,0 +1,11 @@
...
+      if (node.name === 'dataset') {
+        context.report(node, "Due to camel-case transformations, using dataset is not easily greppable. Instead, use el.getAttribute('data-what-ever').")
```
josh: I'd keep this message short and move the *why* to the `docs/` file.

ntwb:
Some docs to go along with the rule would be 👌

lib/rules/no-dataset.js
nakajima: Cool. I made that change in https://github.com/github/eslint-plugin-github/pull/13/commits/2722d990d8751ccaeddb9ba388ddc782f256949c and added some additional test cases.

nakajima: Ok, actually I _really_ did it in https://github.com/github/eslint-plugin-github/pull/13/commits/77a89e0fac80933f565f28097c4521e963f7a3ec.

josh:
@nakajima is this okay to merge or did you want to look into implementing a `fix` function?

nakajima:
I'd say it's good to merge and I can look at the fixer in a follow up.

PR #24, opened 2017-11-14
fix: relax func-style rule; allowArrowFunctions

keithamus:
This alters the func-style rule to allow for arrow functions over function declarations.

This means that `const foo = function(){}` is still invalid code, but it allows for arrow functions which have various tangiable benefits over function declarations.

### DIFF
--- a/lib/configs/recommended.js
+++ b/lib/configs/recommended.js
@@ -4,7 +4,7 @@ module.exports = {
     camelcase: ['error', {properties: 'always'}],
     'constructor-super': 'error',
     eqeqeq: ['error', 'smart'],
-    'func-style': ['error', 'declaration', {'allowArrowFunctions': true}],
+    'func-style': ['error', 'declaration', {allowArrowFunctions: true}],
     'github/no-implicit-buggy-globals': 'error',
     'github/no-sprockets-directives': 'error',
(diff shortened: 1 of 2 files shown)


keithamus:
Individual use cases like that could probably be addressed with the `no-restricted-syntax` custom selector rule; something like:

```js
"no-restricted-syntax": [ "error",
  {
  "selector": "ExportNamedDeclaration > VariableDeclaration > VariableDeclarator > ArrowFunctionExpression",
  "message": "Please only export function declarations, not arrow function expressions"
  }
]
```

josh:
Disregard my previous concern. I thought we still had some legacy grep checks that scanned for exported functions in a naive way. Those don't seem to be around anymore so `export const foo = ` should be fine.

lib/configs/recommended.js
```diff
@@ -4,7 +4,7 @@ module.exports = {
     camelcase: ['error', {properties: 'always'}],
     'constructor-super': 'error',
     eqeqeq: ['error', 'smart'],
-    'func-style': ['error', 'declaration'],
+    'func-style': ['error', 'declaration', { "allowArrowFunctions": true }],
```
josh: Needs single quotes 😁

PR #26, opened 2017-12-08
Enable import/no-anonymous-default-export rule

josh:
This PR enables the [import/no-anonymous-default-export](https://github.com/benmosher/eslint-plugin-import/blob/master/docs/rules/no-anonymous-default-export.md) rule. Ensures all classes and functions exported as `default` have an explicit name for debugging purposes.

This was my other concern with allowing #24 that I couldn't think of at the time. I don't think there's a syntax to use default exports with arrow functions that are also named. (@keithamus correct me if this is something that might be resolved in future spec changes)

``` js
// syntax error
export default const foo = () => {}

// lint error from this rule
export default () => {}

// good
export default function foo() {}
```

### DIFF
--- a/lib/configs/es6.js
+++ b/lib/configs/es6.js
@@ -18,6 +18,7 @@ module.exports = {
     'import/named': 'error',
     'import/namespace': 'error',
     'import/no-absolute-path': 'error',
+    'import/no-anonymous-default-export': 'error',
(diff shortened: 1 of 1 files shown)


keithamus:
> something that might be resolved in future spec changes

There's nothing existing or proposed to change this. Of course we could propose this change ourselves 😉

PR #29, opened 2018-01-17
Unused module detection

josh:
Unused module and export detection is a wishlist item for [eslint-plugin-import](https://github.com/benmosher/eslint-plugin-import). However it's pretty tricky to implement as an eslint rule. The detection really needs to operate over a project as a whole, not on a per file basis which is what eslint expects.

Instead, a new binary `eslint-unused-modules .` is exposed and invoked similar to `eslint .`. The goal is near zero configuration where the new command reuses existing eslint ignore and parser configuration. For an example, if a project uses the `babel-eslint` parser instead of `espree` to parse fancy babel jsx features, it needs to work out of the box.

The one piece of configuration we'll need is for the project to specify it's entry points and some other ignored files. On a simple project we could assume `package.main` is the main entry. But on a larger rails app, we want to be able to specify multiple bundles. It's also useful to ignore tests and other flow junk files.

##

CC: @github/web-systems


koddsson:
Nice! How long does it take to run on something like `github/github`? Can we hook into the eslint reporter to get editor integration for free?

josh:
> Nice! How long does it take to run on something like github/github? Can we hook into the eslint reporter to get editor integration for free?

This is the big issue. The way the editor integrations work—and maintain speed—is that they only recheck one file at a time when that file is modified. Since this is examining an entire project, any single file change could potentially affect the whole dependency graph. So you'd have to rescan all files on any change.

ESLint also doesn't have any built in caching, but this dependency graph could potentially be cached in a custom way persisted to some tmp dir. It'd be a little unusual might be be *fine*. I think the full eslint integration and caching will be out of scope for the initial work. But once we prove this code path out, we can look into that approach.

lib/module-utils.js
```diff
@@ -0,0 +1,171 @@
...
+      for (const identifier of exports) {
+        usedExports.set(identifier, false)
+      }
+    }
+
+    for (const {imports} of results) {
+      for (const [importedPath, importedIdentifiers] of imports) {
+        if (imported.has(importedPath)) {
+          for (const importedIdentifier of importedIdentifiers) {
+            if (importedIdentifier === '*') {
```
mislav: Does this cover the detection that, provided that module `foo` exports `a` and `b`, the unused export gets detected here:

```js
import * as f from "foo"
f.a()
```

lib/module-utils.js
```diff
@@ -0,0 +1,171 @@
...
+exports.parse = function(path, content) {
+  const result = {
+    path,
+    entry: false,
+    imports: new Map(),
+    exports: new Set()
+  }
+
+  const cli = new eslint.CLIEngine()
+  const options = cli.getConfigForFile(path)
```
mislav: How expensive is doing `getConfigForFile` for every individual file? Is this what eslint does normally as well?

lib/module-utils.js
```diff
@@ -0,0 +1,171 @@
...
+function matchPatterns(str, patterns) {
+  for (const pattern of patterns) {
+    if (minimatch(str, pattern)) {
+      return true
+    }
+  }
+  return false
+}
+
+function readFile(path, options) {
```
mislav: You can use `promisify` from `utils` to do `const readFile = promisify(fs.readFile)`

lib/module-utils.js
josh: I think this needs to be revised. It is expensive on our larger internal codebases.

lib/module-utils.js
josh: Oh nice, I didn't realize that was builtin.

lib/module-utils.js
josh: It should, `"*"` was an internal sentinel I used to indicate `import * as f from "foo"` or `import "foo"`.

josh:
I still need to do some refactoring here. Not happy with the performance. I think it could be more tightly integrated into eslint just to simply reuse it's directory traversal and configuration loading facilities.

Maybe it's implemented as a two pass linter where the initial eslint does it's parsing and traversal and outputs a dependency graph. The second pass uses that graph to `context.report` on unused files and exports.

bin/eslint-dependency-graph.js
```diff
@@ -0,0 +1,24 @@
...
+
+const {CLIEngine} = require('eslint')
+
+// TODO: Figure out how to deactive other rules.
+let cli = new CLIEngine({
+  rules: {
+    'github/dependency-graph': 1
+  }
+})
+cli.executeOnFiles(process.argv.slice(2))
```
josh: Step 1 runs eslint on the codebase with the dependency graph rule enabled. The dependency graph rule scans for import/exports and records them as a side effect of running the linter.

bin/eslint-dependency-graph.js
```diff
@@ -0,0 +1,24 @@
...
+
+// TODO: Figure out how to deactive other rules.
+cli = new CLIEngine({
+  rules: {
+    'github/unused-export': 2,
+    'github/unused-module': 2
+  }
+})
```
josh: Step 2 runs the linter again but with new rules enable that check if modules and exports are imported by other files.

bin/eslint-dependency-graph.js
```diff
@@ -0,0 +1,24 @@
...
+
+const report = cli.executeOnFiles(process.argv.slice(2))
+const formatter = cli.getFormatter()
+
+process.stdout.write(formatter(report.results))
```
josh: This prints the results in a standard format you'd expect out of running `eslint`.

bin/eslint-dependency-graph.js
```diff
@@ -0,0 +1,24 @@
+#!/usr/bin/env node
+
+const {CLIEngine} = require('eslint')
+
+// TODO: Figure out how to deactive other rules.
```
josh: Here's an issue, running `CLIEngine` rules with all rules enabled. This is going to slow down the evaluation of all files. We really only need to inject our dependency graph rule. However, we can't disable all configuration. We want to use whatever `parser` and other configuration that would affect how projects parse babel features, flow and jsx.

lib/rules/unused-module.js
```diff
@@ -0,0 +1,12 @@
+const {imported} = require('./dependency-graph')
+
+module.exports = function(context) {
+  return {
+    Program(node) {
+      const {filenames} = imported()
+      if (!filenames.has(context.getFilename())) {
+        context.report(node, 'Module was not imported by any files.')
```
josh: The last gotcha is we need a way to ignore "module entry" points. They won't be imported by any other file. In github's usecase, that's our `*-bootstrap.js` files. They do all the importing but we're going to link to them with a script tag, not another `import`.

bin/eslint-dependency-graph.js
```diff
@@ -0,0 +1,24 @@
...
+    'github/dependency-graph': 1
+  }
+})
+cli.executeOnFiles(process.argv.slice(2))
+
+// TODO: Figure out how to deactive other rules.
+cli = new CLIEngine({
+  rules: {
+    'github/unused-export': 2,
+    'github/unused-module': 2
```
mislav: Is it possible to use `warn/error` keywords instead of magic numbers here?

bin/eslint-dependency-graph.js
mislav: Where are results recorded if a new `CLIEngine` instance is created each time?

bin/eslint-dependency-graph.js
josh: Good call, I believe those work in this context. I'll make that change.

bin/eslint-dependency-graph.js
josh: The side effect state is recorded in a global `dependencyGraph` map.

https://github.com/github/eslint-plugin-github/blob/b88619b0de7e5ea16173769c3954783359b734b4/lib/rules/dependency-graph.js#L5

The second phase uses this helper method to access the state.

https://github.com/github/eslint-plugin-github/blob/b88619b0de7e5ea16173769c3954783359b734b4/lib/rules/dependency-graph.js#L89

PR #30, opened 2018-01-26
add a new `no-innerText` rule

koddsson:
`textContent` seems like a better alternative and we seem to favor that already so if nothing else we can be consistent 😸 

```sh
$ git grep innerText | wc -l
      30
$ git grep textContent | wc -l
     235
```

There are some slight differences between the two and there is probably some code that relies on those differences but there aren't many instances so I can manually check over those to make sure nothing is breaking.

### Reading 📚  
- [`textContent` on MDN](https://developer.mozilla.org/en-US/docs/Web/API/Node/textContent)
- [`innerText` on MDN](https://developer.mozilla.org/en-US/docs/Web/API/Node/innerText)

### DIFF
--- a/lib/rules/no-innerText.js
+++ b/lib/rules/no-innerText.js
@@ -1,0 +1,18 @@
+module.exports = function(context) {
+  return {
+    MemberExpression(node) {
+      if (node.property && node.property.name === 'innerText') {
+        context.report({
+          meta: {
+            fixable: 'code'
+          },
+          node: node.property,
+          message: 'Prefer textContent to innerText',
+          fix(fixer) {
+            return fixer.replaceText(node.property, 'textContent')
+          }
+        })
+      }
+    }
+  }
+}
--- a/tests/no-innerText.js
+++ b/tests/no-innerText.js
@@ -1,0 +1,37 @@
+var rule = require('../lib/rules/no-innerText')
+var RuleTester = require('eslint').RuleTester
+
+var ruleTester = new RuleTester()
+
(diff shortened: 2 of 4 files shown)


lib/rules/no-innerText.js
```diff
@@ -0,0 +1,18 @@
...
+    MemberExpression(node) {
+      if (node.property && node.property.name === 'innerText') {
+        context.report({
+          meta: {
+            fixable: 'code'
+          },
+          node: node.property,
+          message: 'Prefer textContent to innerText',
+          fix(fixer) {
+            return fixer.replaceText(node.property, 'textContent')
```
koddsson: You can run this on a codebase with `--fix` and it will fix it! ✨

tests/no-innerText.js
```diff
@@ -0,0 +1,37 @@
...
+    },
+    {
+      code: 'document.querySelector("js-flash-text").innerText = "bar"',
+      errors: [
+        {
+          message: 'Prefer textContent to innerText',
+          type: 'Identifier'
+        }
+      ],
+      output: 'document.querySelector("js-flash-text").textContent = "bar"'
```
koddsson: This will assert the output if this rule was run with `--fix`.

koddsson:
I want to merge and publish this but I'm not sure if I have the rights on npm to do that.

josh:
> I want to merge and publish this but I'm not sure if I have the rights on npm to do that.

I just added you to our [`@github` org on npm](https://www.npmjs.com/org/github).

PR #31, opened 2018-03-22
add `require-await` rule

koddsson:
It makes sure that you don't have functions marked as `async` when they don't contain any `await` keywords 😄 

### 📚Reading

- [`require-await` rule documentation](https://eslint.org/docs/rules/require-await)

### DIFF
--- a/lib/index.js
+++ b/lib/index.js
@@ -17,6 +17,7 @@ module.exports = {
     'unused-module': require('./rules/unused-module')
   },
   configs: {
+    app: require('./configs/app'),
     browser: require('./configs/browser'),
     es6: require('./configs/es6'),
     flow: require('./configs/flow'),
--- a/lib/configs/app.js
+++ b/lib/configs/app.js
@@ -1,0 +1,7 @@
+module.exports = {
+  plugins: ['github'],
+  rules: {
+    'require-await': 'error'
+  },
+  extends: [require.resolve('./recommended')]
+}
--- a/lib/configs/recommended.js
+++ b/lib/configs/recommended.js
@@ -58,6 +58,7 @@ module.exports = {
     'object-shorthand': ['error', 'always', {avoidQuotes: true}],
     'prefer-promise-reject-errors': 'error',
     'prettier/prettier': 'error',
+    'require-await': 'error'
     'require-yield': 'error',
     'use-isnan': 'error',
     'valid-typeof': 'error'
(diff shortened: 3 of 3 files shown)


koddsson:
@josh @dgraham I created a new config set like we discussed. Is this how you envisioned it? 

I thought the name `app` was a bit generic but couldn't come up with a better one 😄

mislav:
If we're going to differentiate between “app” and “library” sets of rules, then we should also disable `no-then` checks for “library” use-case, since we don't want to force libraries to use `async..await`.

Also, isn't there a valid use-case for `async` functions that don't have `await`, @keithamus? You explained to me that functions can be marked `async` to fulfill an async contract (if they're being passed around as arguments), even though their current implementation might not be async.

lib/index.js
```diff
@@ -17,6 +17,7 @@ module.exports = {
     'unused-module': require('./rules/unused-module')
   },
   configs: {
+    app: require('./configs/app'),
     browser: require('./configs/browser'),
```
dgraham: Since we use the `browser` rule set as the "app" today, my hunch is that we want to add the `require-await` rule to `browser` and introduce a new `lib` or `library` rule set that disables that rule.

keithamus:
> You explained to me that functions can be marked async to fulfill an async contract

Yes an `async` function without any `await`s inside is still an async function. It will always return a Promise, and so can be considered a form type declaration that is built into the language. Additionally consider that `return` also kind of means `await` inside of an async function. To give some examples:

```js
foo = async () => 1 // this is very different but may be desirable to...
foo = () => 1

// If this rule is in place, to get a always-returning-promise-function one must do:
foo = () => Promise.resolve(1)
```

These are perhaps edge cases but also if we look at github-asset-pipeline which relies heavily on these kinds of contracts we can see that sometimes async functions are used without ever using `await`

josh:
I'm slightly in favor of allowing this, but it's not a huge deal either way. I've made use if it in the past to satisfy a contracts (sometimes flow checked) for a function that expected to return a promise but the implementation happened to be sync.

```js
function checker(fn: () => Promise<boolean>) {
  fn.then(result => console.log('ok', result)
}

checker(async function() {
  const data = await fetch("/check")
  return data.ok
})

checker(async function() {
 return confirm("check")
})
```

Though, you could ague the checker function is of poor form as if it was using `await` instead of `then` it would work with the unwrapped value just fine. But I've usually hit this when writing against library code where I didn't have control over `checker`. In general I think avoiding wrapping values in a resolved promise shouldn't be needed. But sometimes you gotta deal with stinky libs.

So I'm whatever here. 

Just curious if there's potential bugs we'd prevent by enabling this today?

koddsson:
To clarify, I just wanted this rule after I was doing some refactoring and accidentally left some `async` keywords on functions after moving all the `await` keywords out of it. Since it's not helping prevent any bugs right now I'm leaning towards closing this PR since it seems to add more complexity and edge cases then it solves.

PR #34, opened 2018-04-11
Separate "app" config

josh:
This PR introduces a breaking change that would move web application specific rules into a `github/app` config set. Other rulesets would assume a reusable library project type.

### DIFF
--- a/bin/eslint-github-init.js
+++ b/bin/eslint-github-init.js
@@ -5,6 +5,7 @@ const fs = require('fs')
 const path = require('path')
 
 const defaults = {
+  project: 'lib',
   env: 'browser',
   flow: true,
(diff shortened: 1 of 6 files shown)


bin/eslint-github-init.js
```diff
@@ -14,12 +15,20 @@ const defaults = {
 const packagePath = path.resolve(process.cwd(), 'package.json')
 if (fs.existsSync(packagePath)) {
   const packageJSON = fs.readFileSync(packagePath, 'utf8')
+  defaults.project = /private/.test(packageJSON) ? 'app' : 'lib'
```
koddsson: This is a neat way to determine if the project is a app or a lib!

PR #38, opened 2018-06-20
Enable more flow rules

koddsson:
In order to encourage more granular type definitions I propose we enable [`flowtype/no-weak-types`](https://github.com/gajus/eslint-plugin-flowtype#eslint-plugin-flowtype-rules-no-weak-types).

It shows a eslint error when someone tries to type a value as `Function` or `Object` rather then the more granular `(string) => boolean` and `{key: string}`


koddsson:
@josh Since you've done most of the versioning it seems, what do you think I should bump the version by here? Technically it's a breaking change? Not sure if we are doing semantic versioning or not?

josh:
Let’s just do a minor bump. This repo will never to be able to follow strict semver as any rule file change will always be considered “breaking” to a degree.

PR #39, opened 2018-06-22
upgrade eslint-plugin-flowtype to latest version

koddsson:
We switched out our own `no-flow-fix-me-comments` rule for the one in `eslint-plugin-flowtype` in #37 but there seems to be a issue with ESLint and `no-flow-fix-me-comments`. Fortunately it was fixed [here](https://github.com/gajus/eslint-plugin-flowtype/pull/338) and has been available since version 2.49.1.

This patch updates the `eslint-plugin-flowtype` package to the latest version (2.49.3).

### DIFF
--- a/package.json
+++ b/package.json
@@ -35,7 +35,7 @@
   "dependencies": {
     "babel-eslint": ">=8.2.0",
     "eslint-config-prettier": ">=2.9.0",
-    "eslint-plugin-flowtype": ">=2.46.0",
+    "eslint-plugin-flowtype": "^2.49.3",
     "eslint-plugin-graphql": ">=2.1.0",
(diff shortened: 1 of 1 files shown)


package.json
```diff
@@ -35,7 +35,7 @@
   "dependencies": {
     "babel-eslint": ">=8.2.0",
     "eslint-config-prettier": ">=2.9.0",
-    "eslint-plugin-flowtype": ">=2.46.0",
+    "eslint-plugin-flowtype": "^2.49.3",
```
keithamus: Maybe keep the range signifier as `>=`?

koddsson: Sure! I just ran `npm install --save eslint-plugin-flowtype@latest`

PR #41, opened 2018-06-27
Add prettier.config.js to files field

shawnbot:
The latest version (1.1.2) doesn't include `prettier.config.js`, which causes instances like this to fail:

```js
// prettier.config.js
module.exports = require('eslint-plugin-github/prettier.config')
```

This adds it to the `files` field in `package.json`. :v:

### DIFF
--- a/package.json
+++ b/package.json
@@ -54,7 +54,8 @@
   },
   "files": [
     "bin/*",
-    "lib/*"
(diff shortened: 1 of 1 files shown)


koddsson:
Thanks for this @shawnbot 💖

PR #42, opened 2018-07-25
Add async currentTarget rule

josh:
Adds rule to prevent error prone access of `event.currentTarget` in async functions.

##

CC: @github/js

### DIFF
--- a/docs/rules/async-currenttarget.md
+++ b/docs/rules/async-currenttarget.md
@@ -1,0 +1,43 @@
+# `event.currentTarget` in an async function
+
+Accessing `event.currentTarget` inside an `async function()` will likely be `null` as `currentTarget` is mutated as the event is propagated.
+
+```js
+// bad
+document.addEventListener('click', async function(event) {
+  // event.currentTarget will be an HTMLElement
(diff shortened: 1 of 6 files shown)


docs/rules/async-currenttarget.md
```diff
@@ -0,0 +1,43 @@
...
+  const url = currentTarget.getAttribute('data-url')
+
+  // call async IIFE
+  ;(async function() {
+    const data = await fetch(url)
+
+    const text = currentTarget.getAttribute('data-text')
+    // ...
+  })()
+})
```
dgraham: Maybe another solution to suggest is extracting the async behavior into a function.

```js
async function fetchData(el) {
  const url = el.getAttribute('data-url')
  const data = await fetch(url)
  const text = el.getAttribute('data-text')
}
document.addEventListener('click', function(event) {
  fetchData(event.currentTarget)
})
```

josh: Yeah, I think that's ideal in practice. Though, solely including that example it's more subtle that the function call argument capture is resolving the issue. I'll add both as examples.

PR #43, opened 2018-07-30
Relax async rules till after first await

josh:
Because async function are invoked synchronously, accessing `event` properties isn't dangerous until the first `await` is called which defers the remainder of the function till after a microtask.

Where before this would be flagged as a lint here, it is actually safe:

```js
on('click', '.foo', async function(event) {
  // this is safe
  event.preventDefault()

  await fetch()

  // this is the issue
  event.preventDefault()
})
```

This change relaxes the linter until after the first `await`.

But it's not bulletproof. You can imagine another case like:

```js
async function doMore(event) {
  // buggy, but won't be flagged by linter
  event.preventDefault()

  await fetchMore()
}

on('click', '.foo', async function(event) {
  await fetch()
  await doMore(event)
})
```

However, this code isn't often written. Really, checking for static symbols like `preventDefault` or `currentTarget` isn't perfect either. To really get this analysis correct, it might have to be done at the flow layer. But I think this trade-off is acceptable.

### DIFF
--- a/lib/rules/async-currenttarget.js
+++ b/lib/rules/async-currenttarget.js
@@ -1,9 +1,14 @@
 module.exports = function(context) {
+  const scopeDidWait = new WeakSet()
+
   return {
+    AwaitExpression() {
+      scopeDidWait.add(context.getScope(), true)
+    },
     MemberExpression(node) {
       if (node.property && node.property.name === 'currentTarget') {
         const scope = context.getScope()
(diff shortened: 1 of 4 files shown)


keithamus:
We could always add an additional lint rule that avoids you passing `event` to other functions - instead suggesting other functions should get the arguments they need rather than a whole event object.

PR #44, opened 2018-08-31
make sure `env.node = true` when user selects node env

koddsson:
Should this rather extend a config like `brower` and `app` do rather then setting it directly?


koddsson:
> Should this rather extend a config like brower and app do rather then setting it directly?

I'm thinking we'd want that so we can also ignore the `no-console` rule?

josh:
> Should this rather extend a config like brower and app do rather then setting it directly?

Yeah, maybe we can have a little node config stub where `no-console` can be allowed. I feel like that should be limited to browsers.

PR #48, opened 2018-09-28
Lint against unescaped html literals

josh:
Pushing developers towards [lit-html](https://github.com/Polymer/lit-html) over unescaped raw HTML literals.

This initial implementation is pretty simple, it's just checking to see if the string starts with an HTML opening tag. I don't know how sophisticated we want to get. Concerned about false positives. Maybe we can start with this dumb check and improve it as we find real world examples that miss the linter.

##

CC: @github/appsec how this will make you 😄

### DIFF
--- a/lib/rules/unescaped-html-literal.js
+++ b/lib/rules/unescaped-html-literal.js
@@ -1,0 +1,25 @@
+module.exports = function(context) {
+  const htmlOpenTag = /<[a-zA-Z]/
+  const message = 'Unescaped HTML literal. Use html`` tag template literal for secure escaping.'
+
+  return {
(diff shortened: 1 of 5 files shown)


gregose:
Very cool! We can add `lit-html` to our Secure Coding Guidelines too.

lib/rules/unescaped-html-literal.js
```diff
@@ -0,0 +1,25 @@
+module.exports = function(context) {
+  const htmlOpenTag = /<[a-zA-Z]/
```
mislav: Concerned about false positives too 🤔 
but let's see.

PR #51, opened 2018-10-05
Add valid camelCase SVG attribute whitelist (and lowercase blacklist)

### LINKED ISSUE #50: github/get-attribute invalidates camelCase SVG attributes
The case of attributes such as `viewBox` needs to be preserved in `setAttribute()` calls. There should probably be a whitelist of SVG attributes (`clipPath`, `preserveAspectRatio`, `patternTransform`, and [units-related attributes](https://www.w3.org/TR/SVG/coords.html#ObjectBoundingBoxUnits)) that don't trigger errors or purport to be fixable.

shawnbot:
This adds an whitelist for SVG attribute names that _must_ be specified in the correct case, such as `clipPath`, `preserveAspectRatio`, and `viewBox`, collected from [`svg-element-attributes`](https://github.com/wooorm/svg-element-attributes/blob/master/index.json). Because we can't reliably detect which element names the code is calling attribute methods on, only whitelist attributes that don't match the `validAttributeName` pattern.

Additionally, a _blacklist_ is created from lowercase variants of all of the camelCase SVG attributes, and the test fails on any of those because they're most likely an error. In other words, `viewBox` is valid, but `viewbox` is not.

Fixes #50.

### DIFF
--- a/lib/rules/get-attribute.js
+++ b/lib/rules/get-attribute.js
@@ -3,7 +3,7 @@ module.exports = function(context) {
   const validAttributeName = /^[a-z][a-z0-9-]*$/
   // these are common SVG attributes that *must* have the correct case to work
   const attributeWhitelist = new Set(['clipPath', 'preserveAspectRatio', 'viewBox'])
-  
+
   function isValidAttribute(name) {
     return attributeWhitelist.has(name) || validAttributeName.test(name)
   }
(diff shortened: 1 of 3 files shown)


josh:
Thanks for fixing this @shawnbot! Hopefully this didn't block anything on your end.

PR #52, opened 2018-10-08
new rule: enforce `name` getter on custom elements

koddsson:
Most minifiers will rename variables, classes and functions to a single letter, or more if the alphabet has been exhausted. This means that any call to `Function.name` will result in unwanted results.

This patch adds a rule that applies to custom elements. The rule enforces the existence of a `static get name()` method that returns the name of the class as a string. Ensuring that any call to `.name` on the custom element will result in the correct name.

### DIFF
--- a/lib/configs/browser.js
+++ b/lib/configs/browser.js
@@ -8,7 +8,8 @@ module.exports = {
     'github/async-preventdefault': 'error',
     'github/get-attribute': 'error',
     'github/no-innerText': 'error',
-    'github/unescaped-html-literal': 'error'
+    'github/unescaped-html-literal': 'error',
+    'github/custom-elements-must-have-name': 'error'
   },
   extends: [require.resolve('./recommended')]
 }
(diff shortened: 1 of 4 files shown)


koddsson:
> This is great!
> 
> I was also considering a potential solution to create (or extend the existing) Babel transform for custom element classes to add a static `name` property. That way, instead of linting it, we would generate it.
> 
> However, if we ever drop [babel-plugin-transform-custom-element-classes](https://github.com/github/babel-plugin-transform-custom-element-classes/blob/master/lib/index.js) because shipping ES5-compatible syntax would no longer be needed, then the Babel-based workaround for this would expire too. So maybe it's better in the long run to take the linter approach.

Thanks @mislav. I think you saw on slack already but @github/web-systems seemed to lean more towards using a babel plugin. Luckily it was easy to port and I just published https://github.com/github/babel-plugin-ensure-name-for-custom-elements ✨ 

I'll close this when https://github.com/github/include-fragment-element/pull/39 gets merged.

koddsson:
> I'll close this when [github/include-fragment-element#39](https://github.com/github/include-fragment-element/pull/39) gets merged.

Actually, we are probably not going this direction in any case so I'm fine with closing this now.

PR #55, opened 2018-11-30
[WIP] Typescript

koddsson:
Add support for linting typescript projects.

### DIFF
--- a/bin/eslint-github-init.js
+++ b/bin/eslint-github-init.js
@@ -76,6 +76,18 @@ inquirer.prompt(questions).then(answers => {
     eslintrc.extends.push('plugin:github/typescript')
    
     // TODO: Check if tsconfig.json exists, generate it if it doesn't.
+    const tsconfigPath = path.resolve(process.cwd(), 'tsconfig.json')
+    if (!fs.existsSync(tsconfig)) {
+      const tsconfigDefaults = {
+        "compilerOptions": {
+          "target": "es5",
+          "module": "commonjs",
+          "strict": true,
+          "esModuleInterop": true
+        }
+      }
+      fs.writeFileSync(tsconfig, JSON.stringify(tsconfigDefaults, null, '  '), 'utf8')
+    }
     
     const tslintPath = path.resolve(process.cwd(), 'tslint.json')
     const tslintrc = fs.existsSync(tslintPath) ? JSON.parse(fs.readFileSync(tslintPath, 'utf8')) : {
--- a/bin/eslint-github-init.js
+++ b/bin/eslint-github-init.js
@@ -88,7 +88,7 @@ inquirer.prompt(questions).then(answers => {
       rulesDirectory: []
     }
     if (tslintrc.extends[tslintrc.extends.length - 1] !== 'tslint-config-prettier') {
-      tslint.extends.push('tslint-config-prettier')
+      tslintrc.extends.push('tslint-config-prettier')
     }
     fs.writeFileSync(path.resolve(process.cwd(), 'tslint.json'), JSON.stringify(tslintrc, null, '  '), 'utf8')
   }
--- a/bin/eslint-github-init.js
+++ b/bin/eslint-github-init.js
@@ -77,7 +77,7 @@ inquirer.prompt(questions).then(answers => {
    
     // TODO: Check if tsconfig.json exists, generate it if it doesn't.
     const tsconfigPath = path.resolve(process.cwd(), 'tsconfig.json')
-    if (!fs.existsSync(tsconfig)) {
+    if (!fs.existsSync(tsconfigPath)) {
       const tsconfigDefaults = {
         "compilerOptions": {
           "target": "es5",
@@ -86,7 +86,7 @@ inquirer.prompt(questions).then(answers => {
           "esModuleInterop": true
         }
       }
-      fs.writeFileSync(tsconfig, JSON.stringify(tsconfigDefaults, null, '  '), 'utf8')
+      fs.writeFileSync(tsconfigPath, JSON.stringify(tsconfigDefaults, null, '  '), 'utf8')
     }
     
     const tslintPath = path.resolve(process.cwd(), 'tslint.json')
--- a/bin/eslint-github-init.js
+++ b/bin/eslint-github-init.js
@@ -81,17 +81,10 @@ inquirer.prompt(questions).then(answers => {
       extends: [
         "tslint-config-prettier"
       ],
-      rulesDirectory: [
-        "tslint-plugin-prettier"
-      ],
-      jsRules: {
-        prettier: true
-      },
-      rules: {
-        prettier: true
-      }
+      jsRules: {},
+      rules: {},
+      rulesDirectory: []
     }
-
     fs.writeFileSync(path.resolve(process.cwd(), 'tslint.json'), JSON.stringify(tslintrc, null, '  '), 'utf8')
   }
   if (answers.react) eslintrc.extends.push('plugin:github/react')
--- a/bin/eslint-github-init.js
+++ b/bin/eslint-github-init.js
@@ -72,7 +72,21 @@ inquirer.prompt(questions).then(answers => {
   }
 
   if (answers.typeSystem === 'flow') eslintrc.extends.push('plugin:github/flow')
-  if (answers.typeSystem === 'typescript') eslintrc.extends.push('plugin:github/typescript')
+  if (answers.typeSystem === 'typescript') {
+    eslintrc.extends.push('plugin:github/typescript')
+    
+    // TODO: Check if tslint exists, read it if exists and make sure `tslint-config-prettier` is the last string.
+    const tslintrc = {
+      defaultSeverity: "error",
+      extends: [
+        "tslint-config-prettier"
+      ],
+      jsRules: {},
+      rules: {},
+      rulesDirectory: []
+    }
+    fs.writeFileSync(path.resolve(process.cwd(), 'tslint.json'), JSON.stringify(tslintrc, null, '  '), 'utf8')
+  }
   if (answers.react) eslintrc.extends.push('plugin:github/react')
   if (answers.relay) eslintrc.extends.push('plugin:github/relay')
 
--- a/bin/eslint-github-init.js
+++ b/bin/eslint-github-init.js
@@ -74,9 +74,11 @@ inquirer.prompt(questions).then(answers => {
   if (answers.typeSystem === 'flow') eslintrc.extends.push('plugin:github/flow')
   if (answers.typeSystem === 'typescript') {
     eslintrc.extends.push('plugin:github/typescript')
+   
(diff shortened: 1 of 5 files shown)


koddsson:
Whoops! Opened this on my fork instead of a branch on remote. I'll switch over so CI will run on the PR.

bin/eslint-github-init.js
```diff
@@ -68,14 +71,49 @@ inquirer.prompt(questions).then(answers => {
...
   if (answers.react) eslintrc.extends.push('plugin:github/react')
   if (answers.relay) eslintrc.extends.push('plugin:github/relay')
 
   fs.writeFileSync(path.resolve(process.cwd(), '.eslintrc.json'), JSON.stringify(eslintrc, null, '  '), 'utf8')
 
   const prettierConfig = []
-  if (answers.flow) prettierConfig.push('/* @flow */')
+  if (answers.typeSystem === 'flow') prettierConfig.push('/* @flow */')
+
+  // TODO: Set arrow parens to true if typescript
```
koddsson: I don't think this is needed. I think `tslint-config-prettier` does this already.

bin/eslint-github-init.js
```diff
@@ -68,14 +71,49 @@ inquirer.prompt(questions).then(answers => {
...
+    if (!fs.existsSync(tsconfigPath)) {
+      const tsconfigDefaults = {
+        compilerOptions: {
+          target: 'es5',
+          module: 'commonjs',
+          strict: true,
+          esModuleInterop: true
+        }
+      }
+      fs.writeFileSync(tsconfigPath, JSON.stringify(tsconfigDefaults, null, '  '), 'utf8')
```
koddsson: We don't need to create a `tsconfig.json` like we have to create `tslint.json` but I figured it would lead to less friction for developers.

Note that we don't create `.flowconfig` for flow.

bin/eslint-github-init.js
```diff
@@ -68,14 +71,49 @@ inquirer.prompt(questions).then(answers => {
     eslintrc.extends.push('plugin:github/browser')
   }
 
-  if (answers.flow) eslintrc.extends.push('plugin:github/flow')
+  if (answers.typeSystem === 'flow') eslintrc.extends.push('plugin:github/flow')
+  if (answers.typeSystem === 'typescript') {
+    eslintrc.extends.push('plugin:github/typescript')
+
+    // TODO: Keep tsconfig.json and tslint.json as files in the lib and copy it over, rather then keeping it here inline.
```
koddsson: TODO

koddsson: Actually the more I look at this, I think this is correct. I don't think we keep any other config as a file currently so I think I'll follow that path.

koddsson: Oh we do for prettier!

bin/eslint-github-init.js
```diff
@@ -68,14 +71,51 @@ inquirer.prompt(questions).then(answers => {
...
+  if (answers.typeSystem === 'typescript') {
+    eslintrc.extends.push('plugin:github/typescript')
+
+    // Create a `tsconfig.json` if one doesn't exits. Equivalent to running `tsc --init`.
+    const tsconfigPath = path.resolve(process.cwd(), 'tsconfig.json')
+    if (!fs.existsSync(tsconfigPath)) {
+      const tsconfigDefaults = {
+        compilerOptions: {
+          target: 'es5',
+          module: 'commonjs',
```
keithamus: Is this correct? We don't use `commonjs` for authoring or artefacts. Should this be `esnext`?

bin/eslint-github-init.js
```diff
@@ -68,14 +71,51 @@ inquirer.prompt(questions).then(answers => {
...
+  if (answers.typeSystem === 'flow') eslintrc.extends.push('plugin:github/flow')
+  if (answers.typeSystem === 'typescript') {
+    eslintrc.extends.push('plugin:github/typescript')
+
+    // Create a `tsconfig.json` if one doesn't exits. Equivalent to running `tsc --init`.
+    const tsconfigPath = path.resolve(process.cwd(), 'tsconfig.json')
+    if (!fs.existsSync(tsconfigPath)) {
+      const tsconfigDefaults = {
+        compilerOptions: {
+          target: 'es5',
```
keithamus: I think our browser support matrix could afford this to be `es6`

bin/eslint-github-init.js
```diff
@@ -68,14 +71,51 @@ inquirer.prompt(questions).then(answers => {
...
+          jsRules: {},
+          rules: {},
+          rulesDirectory: []
+        }
+
+    // If `tslint-config-prettier` isn't the last string in the `extends` array.
+    if (tslintrc.extends[tslintrc.extends.length - 1] !== 'tslint-config-prettier') {
+      tslintrc.extends.push('tslint-config-prettier')
+    }
+    fs.writeFileSync(path.resolve(process.cwd(), 'tslint.json'), JSON.stringify(tslintrc, null, '  '), 'utf8')
```
keithamus: A nice to have here might be to bring in [`detect-indent`](http://npm.im/detect-indent) so that an init doesn't rewrite my files and chose its own indent style.

keithamus: I personally prefer passing in an int instead of a string to `JSON.stringify` as I find it reads clearer:

```suggestion
    fs.writeFileSync(path.resolve(process.cwd(), 'tslint.json'), JSON.stringify(tslintrc, null, 2), 'utf8')
```

Each to their own of course

koddsson: Reading the `detect-indent` API it occurred to me that I can't pass in a number since the indent can be a tab so I pass the `indent` property directly into the JSON stringify function.

bin/github-lint.js
```diff
@@ -22,7 +22,17 @@ function execFile(command, args) {
 
   const packageJson = fs.existsSync('package.json') ? require(path.join(process.cwd(), 'package.json')) : {}
 
-  commands.push(['eslint', ['--report-unused-disable-directives', '.']])
+  let eslintOptions = ['--report-unused-disable-directives', '.']
+
+  if (fs.existsSync('tsconfig.json')) {
+    eslintOptions = eslintOptions.concat(['--ext', '.ts'])
```
koddsson:
```suggestion
    eslintOptions = eslintOptions.concat(['--ext', '.ts,.tsx'])
```

bin/eslint-github-init.js
```diff
@@ -68,14 +72,59 @@ inquirer.prompt(questions).then(answers => {
...
+    // Create a `tsconfig.json` if one doesn't exits. Equivalent to running `tsc --init`.
+    const tsconfigPath = path.resolve(process.cwd(), 'tsconfig.json')
+    if (!fs.existsSync(tsconfigPath)) {
+      const tsconfigDefaults = {
+        compilerOptions: {
+          target: 'ES2015',
+          module: 'commonjs',
+          strict: true,
+          esModuleInterop: true
+        }
```
koddsson: We should set `"jsx": "react"` when user is using react.

koddsson: Also note that when we do that and read the `tsconfig.json` that it's a JSON5 file so we probably need to import a lib to parse it correctly.

bin/eslint-github-init.js
keithamus: From what I can read I guess we want `ESNext` as that supports `import()` as well as `import`

bin/eslint-github-init.js
```diff
@@ -16,7 +17,9 @@ const packagePath = path.resolve(process.cwd(), 'package.json')
 if (fs.existsSync(packagePath)) {
   const packageJSON = fs.readFileSync(packagePath, 'utf8')
   defaults.project = /private/.test(packageJSON) ? 'app' : 'lib'
-  defaults.flow = /flow/.test(packageJSON)
+  if (/typescript/.test(packageJSON)) {
```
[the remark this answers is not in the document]
keithamus: An easy solution to that is just to add word boundaries

```suggestion
    if (/\btypescript\b/.test(packageJSON)) {
```

Although I do agree - I'd sooner see a test something like:

```suggestion
    if ('typescript' in packageJSON.dependencies || 'typescript' in packageJSON.devDependencies) {
```

bin/eslint-github-init.js
```diff
@@ -68,14 +72,59 @@ inquirer.prompt(questions).then(answers => {
...
   }
 
-  if (answers.flow) eslintrc.extends.push('plugin:github/flow')
+  if (answers.typeSystem === 'flow') eslintrc.extends.push('plugin:github/flow')
+  if (answers.typeSystem === 'typescript') {
+    eslintrc.extends.push('plugin:github/typescript')
+
+    // Create a `tsconfig.json` if one doesn't exits. Equivalent to running `tsc --init`.
+    const tsconfigPath = path.resolve(process.cwd(), 'tsconfig.json')
+    if (!fs.existsSync(tsconfigPath)) {
```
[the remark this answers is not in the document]
koddsson: It might be, I'm not too sure about this but we might have to make sure that it is configured in a certain way. We could of course just skip this and document it instead.

koddsson:
I think I took this out of WIP status too soon as I wasn't actually running the linter correctly against my test project and after I made some changes I started getting _a lot_ of errors!

Fixing those there and then I'll bring the changes I make to the linter there into this PR.

bin/eslint-github-init.js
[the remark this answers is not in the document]
mislav: Agreed with @keithamus: I'd rather test `dependencies/devDependencies` directly, or not perform any magic configuration at all. 👍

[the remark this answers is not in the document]
koddsson: Sounds good. I'm gonna work on this PR quite a bit more to make sure it's rock solid. I might do those changes in a separate PR and merge that first so I can keep the diff for this change small.

josh:
This is probably just me being ignorant of tslint and the TypeScript ecosystem, so correct me if I'm wrong.

But how much overlap is there going to be when writing rules that are somehow both compatible with ES and TS? Would we be better off with another project `tslint-plugin-github` that best integrated with that toolchain?

koddsson:
> This is probably just me being ignorant of tslint and the TypeScript ecosystem, so correct me if I'm wrong.

Not at all!

> But how much overlap is there going to be when writing rules that are somehow both compatible with ES and TS? Would we be better off with another project `tslint-plugin-github` that best integrated with that toolchain?

With this PR we parse the typescript code with the eslint typescript parser so we should be able to apply eslint rules to typescript code as far as I know so any ESLint rule should be a rule in Typescript code as well.

If we want to simplify we could definitely do a separate package but I figured since `typescript-eslint-parser` exists we could reuse our eslint config.

koddsson:
@github/web-systems I think this is ready for a re-review. I can still extract this out and have a separate TSLint package if we want but I'm liking how this turned out.

koddsson:
I haven't heard back and this has a ✅so I'm gonna merge on green 🕺

PR #62, opened 2019-01-17
Add eslint-disable-next-line instead of eslint-disable comments

muan:
With #54, global disable comments are no longer allowed. This changes so `eslint-ignore-errors filepath.js` would add `eslint-disable-next-line` comments instead.

This also handles the case where a `eslint-disable-next-line` comment already exist. It will append the rule to the existing comment instead of adding another comment.

This is what https://github.com/github/github/pull/105648 used! 🙌 

cc @josh

### DIFF
--- a/bin/eslint-ignore-errors.js
+++ b/bin/eslint-ignore-errors.js
@@ -1,5 +1,5 @@
 #!/usr/bin/env node
-// Disables eslint rules in a JavaScript file with global comments. This is
+// Disables eslint rules in a JavaScript file with next-line comments. This is
 // useful when introducing a new rule that causes many failures. The comments
 // can be fixed and removed at while updating the file later.
 //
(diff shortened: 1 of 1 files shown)


bin/eslint-ignore-errors.js
```diff
@@ -13,14 +13,43 @@ const execFile = require('child_process').execFile
...
+      fs.writeFileSync(filename, jsLines.join('\n'), 'utf8')
     }
   })
 })
+
+function isDisableComment(line) {
+  return line.match(/\/\* eslint-disable-next-line .+\*\//)
+}
+
+function leftPad(line) {
```
koddsson: You could always just install this from `npm` :trollface:

bin/eslint-ignore-errors.js
```diff
@@ -13,14 +13,43 @@ const execFile = require('child_process').execFile
...
+  const spaceMatch = line.match(/^\s+/g)
+  const spaces = spaceMatch ? spaceMatch[0].length : 0
+  return new Array(spaces).fill(' ').join('')
```
keithamus: You can use `String.prototype.repeat` here

```suggestion
  return ' '.repeat(spaces)
```

bin/eslint-ignore-errors.js
```diff
@@ -13,14 +13,43 @@ const execFile = require('child_process').execFile
...
     }
   })
 })
+
+function isDisableComment(line) {
+  return line.match(/\/\* eslint-disable-next-line .+\*\//)
+}
+
+function leftPad(line) {
+  const spaceMatch = line.match(/^\s+/g)
```
keithamus: If you change this to `*` instead of plus then you can skip the optional check, so this whole function becomes:

```js
const leftPad = line => ' '.repeat(line.match(/^\s*/g)[0].length)
```

PR #71, opened 2019-08-20
Add Flow to TypeScript transitioning rule

josh:
Adds rule to enforce either `@flow` or `@ts-check` on a file. Projects transitioning to TypeScript should enable this rule and disable the [flow ruleset](https://github.com/github/eslint-plugin-github/blob/master/lib/configs/flow.js).


josh:
Think `master` might be red. 🤔

PR #77, opened 2019-10-25
Update packages and set `>=` as version operator.

koddsson:
This should mean that we don't have to bump these and cut a release. The applications using the plugin should be able to update to whatever version of these sub-dependencies that they want.

Should there even be a `package-lock.json`?


package.json
```diff
@@ -33,25 +33,24 @@
     "node": ">=8.11.1"
   },
   "dependencies": {
-    "@typescript-eslint/eslint-plugin": "^1.3.0",
-    "@typescript-eslint/parser": "^1.3.0",
```
dgraham: The motivation for changing these two plugin versions is to remove the warning while building remote-input-element.

```
$ cd github/remote-input-element 
$ npm run build
…
> eslint --report-unused-disable-directives . --ext .js,.ts,.tsx
=============

WARNING: You are currently running a version of TypeScript which is not officially supported by typescript-estree.

You may find that it works just fine, or you may not.

SUPPORTED TYPESCRIPT VERSIONS: >=3.2.1 <3.6.0

YOUR TYPESCRIPT VERSION: 3.6.4

Please only submit bug reports when using the officially supported version.

=============

> tsc --noEmit
```

PR #78, opened 2019-11-06
Change default setting from a string to a object

koddsson:
After #77 we make consumers of this lib install the latest versions of all the dependencies so when packages make breaking changes it will break in those applications.

One of those breaking changes come from `@typescript-eslint/eslint-plugin` where the array type option was changed from a simple string to an options object.

This PR changes the array type option from a `'array-simple'` string to a `{default: 'array-simple'}` option object.

The breaking change was made in https://github.com/typescript-eslint/typescript-eslint/pull/654 and you can read the docs for the rule [here](https://github.com/typescript-eslint/typescript-eslint/blob/master/packages/eslint-plugin/docs/rules/array-type.md)

### DIFF
--- a/lib/configs/typescript.js
+++ b/lib/configs/typescript.js
@@ -5,7 +5,7 @@ module.exports = {
   },
   plugins: ['@typescript-eslint', 'github'],
   rules: {
-    '@typescript-eslint/array-type': ['error', 'array-simple'],
+    '@typescript-eslint/array-type': ['error', {default: 'array-simple'}],
     '@typescript-eslint/no-use-before-define': 'off',
     '@typescript-eslint/explicit-member-accessibility': 'off',
     '@typescript-eslint/explicit-function-return-type': 'off'


koddsson:
> Is it possible to ensure backwards compatibility here? Do we need to detect what version is being used and emit the right option?

I'm not sure. We can't read the package.json of this package since it [wants version `>= 2.5.0`](https://github.com/github/eslint-plugin-github/blob/6d821c9354b5e2f9a271e9d47589e225079ef76d/package.json#L36-L37).

We would probably need to find where `@typescript-eslint/parsers` is installed and read it's `package.json` to figure out it's version but that seems fragile to me.

koddsson:
Talked with @keithamus on slack, and we actually don't need backwards compatibilty with `@typescript-eslint/eslint-plugin@1` since we require version `>= 2.5.0` after we merged #77.

PR #79, opened 2019-11-25
Install `v8-compile-cache`

koddsson:
The tests will fail locally on node < 9 claiming that `v8-compile-cache` is missing. This PR just installs that module.


koddsson:
I expected https://github.com/github/eslint-plugin-github/pull/81 to fail with node > 8 but it seems to run fine. Maybe it's only a thing on my macOS? I'll add a macOS to the matrix and see if that fails it.

koddsson:
Hmm, maybe it's not macOS and just my machine? 🤔

koddsson:
@keithamus removed his `node_modules/` folder and everything worked as expected so there's probably some `npm` issue that can be solved with `rm -rf node_modules/` here.

PR #84, opened 2019-12-06
fix: Consistent max buffer value across Node versions.

MaximDevoir:
Before Node v12, `maxBuffer` defaulted to 200kb. Node v12+ defaults to 1024kb.

Setting the `maxBuffer` value allows for a more consistent cross-version Node experience.

### DIFF
--- a/bin/github-lint.js
+++ b/bin/github-lint.js
@@ -9,7 +9,7 @@ const path = require('path')
 
 function execFile(command, args) {
   return new Promise(resolve => {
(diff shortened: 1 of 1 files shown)


MaximDevoir:
On large projects that are just beginning to switch to the rules of `eslint-plugin-github`, using the `github-lint` tool would crash because there were too many lint errors/warnings. So this PR also raises the buffer for Node v8 and Node v10 environments to match Node v12+ buffer value.

PR #85, opened 2019-12-06
feat: Run ESLint with colors in supported environments

MaximDevoir:
My guess is ESLint is disabling color formatting because it's unable to detect a TTY when invoked through `execFile()`.

Comparision

lint-with-no-color

lint-with-color

### DIFF
--- a/bin/github-lint.js
+++ b/bin/github-lint.js
@@ -22,7 +22,7 @@ function execFile(command, args) {
 
   const packageJson = fs.existsSync('package.json') ? require(path.join(process.cwd(), 'package.json')) : {}
 
-  let eslintOptions = ['--report-unused-disable-directives', '.']
+  let eslintOptions = ['--color', '--report-unused-disable-directives', '.']
 
   const isTypeScriptProject = fs.existsSync('tsconfig.json')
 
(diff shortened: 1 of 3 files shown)


koddsson:
I like this but this would get mangled when running in some CI systems that don't support color.

Maybe we can use [`supports-color`](https://github.com/chalk/supports-color) to detect if the terminal supports color or not?

MaximDevoir:
Okay, I'll work on getting that integrated.

MaximDevoir:
I looked through the [eslint](https://github.com/eslint/eslint) repo and it seems that they are using a very limited set of colors. Therefore, to run ESLint with the `--color` flag, the environment only needs to support basic colors (16 colors).

From what I gathered, ESLint uses these colors:

- red
- yellow
- green
- blue
- bold - from [`chalk.bold()`](https://github.com/chalk/chalk#modifiers)
- dim - from [`chalk.dim()`](https://github.com/chalk/chalk#modifiers)

A basic 16-color environment is enough to support the `--colors` flag and the [modifiers]((https://github.com/chalk/chalk#modifiers)) are widely available.

Issue #86, opened 2020-01-28
Why is `forEach` discouraged?


zeke:
I am using this linter in a project and it raised an issue about code like this:

```js
apples
  .filter(apple => !apple.bad)
  .forEach(doSomething)
```

A `for` loop would have been messy in this context as the `forEach` had a number of `const` declarations. I ended up working around it by changing the `forEach` to `map`.

What is the motivation for this rule? My guess is that `forEach` doesn't play nicely with async functions, but I'm not sure. The current doc doesn't provide any explanation of the rationale: https://github.com/github/eslint-plugin-github/blob/master/docs/rules/array-foreach.md

cc @github/product-docs-engineering 
cc @josh @koddsson @keithamus

koddsson:
This rule was added before my time at GitHub but I think was added to steer developers away from converting `NodeList` to an array just to be able to call `forEach` on it.

```js
// Doesn't work since `document.querySelectorAll` returns a NodeList and not an Array.
document.querySelectorAll('.js-some-element').forEach(..)

// Works since we've converted the NodeList to an Array.
Array.from(document.querySelectorAll('.js-some-element')).forEach(..)
```

Since we work a lot with the `.querySelectorAll` method on github.com we operate a lot on `NodeList` and showing developers how to iterate over it rather than converting it to an array first is probably the intent here.

It has the drawback of giving you false positives when working with non `NodeList` variables though like you've experienced.

ldeluigi:
Please fix this rule: for .. of can't be used to iterate with index.

PR #87, opened 2020-01-29
docs: describe Array#forEach anti patterns

keithamus:
Closes https://github.com/github/eslint-plugin-github/issues/86.

This describes in more detail why we discourage `forEach`, instead preferring `for...of`.

/cc @koddsson @zeke

### DIFF
--- a/docs/rules/array-foreach.md
+++ b/docs/rules/array-foreach.md
@@ -50,6 +50,8 @@ The problem we now have is that we're iterating multiple times over the items in
  }
 ```
 
+Chaning isn't always necessarily bad. Chaining can advertise a series of transformations that are independant from one another, and therefore aid readability. Additionally, sometimes the "goto-style" behaviour of `continue` in for loops can hamper readability. For small Arrays, performance is not going to be of concern, but caution should be applied where there is a potentially unbounded Array (such as iterating over a fetched users list) as performance can easily become a bottleneck when unchecked.
+
 ### Hiding Intent
 
 The `forEach` method passes more than just the current item it is iterating over. The signature of the `forEach` callback method is `(cur: T, i: Number, all: []T) => void` and it can _additionally_ override the `receiver` (`this` value), meaning that often the _intent_ of what the callback does is hidden. To put this another way, there is _no way_ to know what the following code operates on without reading the implementation: `forEach(polishApple)`. 
--- a/docs/rules/array-foreach.md
+++ b/docs/rules/array-foreach.md
@@ -23,7 +23,7 @@ Here's a summary of why `forEach` is disallowed, and why we prefer `for...of` fo
  - `for` loops are considered "more readable" and have **clearer intent**.
  - `for...of` loops offer the **most flexibility** for iteration (especially vs `Array.from`).
 
-For more detail, here is a break down of each of those points:
+For more detail, here is a breakdown of each of those points:
 
 ### Layering of bad practices
 
--- a/docs/rules/array-foreach.md
+++ b/docs/rules/array-foreach.md
@@ -14,6 +14,60 @@ for (const el of els) {
 }
 ```
 
+## Why disallow `forEach`
+
+Here's a summary of why `forEach` is disallowed, and why we prefer `for...of` for almost any use-case of `forEach`:
+
+ - Allowing `forEach` encourages **layering of "bad practices"**, such as using `Array.from()` (which is less performant than using `for...of`).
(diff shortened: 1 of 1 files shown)


muan:
I believe @zeke was also right with regards to async functions in that `forEach` is blocking while `for...of` isn't. I vaguely remember that being the reason. cc @dgraham

dgraham:
```js
async function doSomething(x) { await ... }

// Probably doesn't do what you expect.
items.forEach(item => { await doSomething(item })

// vs

// Blocks the loop as expected.
for (const item of items) {
  await doSomething(item)
}
```

keithamus:
This is a syntax error
```js
items.forEach(item => { await doSomething(item) })
````

This will `await`, but the async function itself is not `await`ed because `forEach` does nothing with the return value
```js
items.forEach(async item => { await doSomething(item) })
```

This is how you'd have to refactor this to be async:
```js
await Promise.all(items.map(async item => { await doSomething(item) }))
```

I'll add this as an example in the Flexibility heading.

docs/rules/array-foreach.md
```diff
@@ -14,6 +14,60 @@ for (const el of els) {
...
+
+Often when using a method like `forEach` - when coming back to add new code, let's say to filter certain elements from the Array before operating on them, a developer is implicitly encouraged to use Array's method chaining to achieve this result. For example if we wanted to filter out bad apples from an Array of Apples, if the code already uses `forEach`, then its a simple addition to add `filter()`:
+
+```diff
+ apples
++  .filter(apple => !apple.bad)
+    .forEach(polishApple)
+```
+
+The problem we now have is that we're iterating multiple times over the items in a collection. Using `forEach` to begin with is what encouraged the chaining, if this were a `for` loop then the equivalent behaviour would be to use 2 `for` loops, which a developer is far less likely to write, so the `for` loop instead encourages an imperative style `continue`, keeping within a single set of iterations:
```
dgraham: I don't use forEach because it's purely for side effects, so a for loop avoids creating a closure and advertises it's just doing something for the side effect.

However, I prefer the functional style afforded by map and filter and will happily chain them. This avoids the goto behavior of `continue` and `next` and advertises a series of transformations that are independent of one another. 

Coming back to chained code later, it's easy to insert and remove new transformations. With a for loop, all intent is mixed together and cannot be easily changed.

keithamus: How do you suggest we change this paragraph to reflect that?

dgraham: I don't have an issue with the advice, just noting that personal preference will vary here and loop control flow is a sign that the loop is responsible for more than one thing.

A for loop that needs to both filter and map items can be structured like this to avoid control flow.

```js
function mapItems(items) {
  const results = []
  for (const item of items) {
    if (isSomePredicate(item) {
      results.push(doSomething(item))
    }
  }
  return results
}
const results = mapItems(items)
```

Pulling the loop body into a named function documents what it does and isolates its local variable state from the surrounding loop code. So it has the same benefits as my personal preference for:

```js
const results = items.filter(isSomePredicate).map(doSomething)
```

docs/rules/array-foreach.md
```diff
@@ -14,6 +14,89 @@ for (const el of els) {
...
+## Why disallow `forEach`
+
+Here's a summary of why `forEach` is disallowed, and why we prefer `for...of` for almost any use-case of `forEach`:
+
+ - Allowing `forEach` encourages **layering of "bad practices"**, such as using `Array.from()` (which is less performant than using `for...of`).
+ - When more requirements are added on, `forEach` typically gets **chained** with other methods like `filter` or `map`, causing multiple iterations over the same Array. Encouraging `for` loops discourages chaining and encourages single-iteration logic (e.g. using a `continue` instead of `filter`).
+ - `for` loops are considered "more readable" and have **clearer intent**.
+ - `for...of` loops offer the **most flexibility** for iteration (especially vs `Array.from`).
+
+For more detail, here is a break down of each of those points:
```
zeke:
```suggestion
For more detail, here is a breakdown of each of those points:
```

docs/rules/array-foreach.md
zeke: Not a blocker for shipping this PR, but I just want to echo @dgraham's point that functional-style chaining "advertises a series of transformations that are independent of one another". In general, I think the readability of this chaining is often worth the (usually small) performance impact of doing numerous loops.

keithamus: I've updated to add some clarifications based on the comments.

PR #90, opened 2020-03-12
Major update

koddsson:
Removing a ton of stuff I don't think we need any more as well as trying to simplify the whole repo, hoping to eventually get down to a list of rules and their documentation.

I'm down to reverting any of these changes if you think we should still keep it.

This is a first pass and I want to do more after this is merged.

### DIFF
--- a/package.json
+++ b/package.json
@@ -40,8 +40,7 @@
   },
   "files": [
     "bin/*",
-    "lib/*",
-    "prettier.config.js"
+    "lib/*"
   ],
   "devDependencies": {
     "eslint": ">=6.5.1",


package.json
```diff
@@ -29,45 +24,26 @@
     "url": "https://github.com/github/eslint-plugin-github/issues"
   },
   "homepage": "https://github.com/github/eslint-plugin-github#readme",
-  "engines": {
-    "node": ">=8.11.1"
-  },
```
koddsson: As far as I can tell, this isn't used for anything.

koddsson:
One thing we could do is create a new `eslint-config-github` repo and keep the current repo as is. That would mean that this repo can just be deprecated and archived.

koddsson:
> One thing we could do is create a new `eslint-config-github` repo and keep the current repo as is. That would mean that this repo can just be deprecated and archived.

Apparently you need to name the package `plugin` in order for you to define your own rules. With that in mind I might just break this PR into smaller PRs and keep it in the same repo. I'll bump the major version once I got all those PRs out and maaayyybe rename the package to something like `@github/eslint-plugin`.

koddsson:
image

There's more stuff to do for 4.0.0. I'll open up a tracking issue for those outstanding issues.

PR #98, opened 2020-03-19
Rename app config

koddsson:
The description of the `app` config is:

>Recommended rules when writing a browser application.

Even though we have a `browser` config already. This PR renames the `app` config, the `internal` config and makes sure that it only loads rules that internal github applications would want.

Also removed the `es6` part of the docs because I missed them in #96


bin/eslint-github-init.js
```diff
@@ -47,7 +47,7 @@ const questions = [
   }
 ]
 
-inquirer.prompt(questions).then(answers => {
+for (const answers of inquirer.prompt(questions)) {
```
keithamus: does this need an `await`?

bin/github-lint.js
```diff
@@ -17,48 +17,50 @@ function execFile(command, args) {
...
-    process.stderr.write(`\nCommand failed: ${nonzero}\n`)
-    process.exit(nonzero)
-  }
-})().catch(error => {
+    const nonzero = codes.find(code => code !== 0)
+    if (nonzero) {
+      process.stderr.write(`\nCommand failed: ${nonzero}\n`)
+      process.exit(nonzero)
+    }
+  })
```
keithamus: This doesn't look right. There needs to be parens to call this function

bin/eslint-github-init.js
koddsson: Oh shoot, this was supposed to be a `await/async` conversion and not a `for-of` one.

PR #104, opened 2020-04-01
Change typescript rules

koddsson:
eae37ab

`@typescript-eslint` already includes a [`no-unused-vars`](https://github.com/typescript-eslint/typescript-eslint/blob/master/packages/eslint-plugin/docs/rules/no-unused-vars.md) rule so we use that one instead.

be71c95

I think we want to just allow people to name their interfaces whatever they want.

### DIFF
--- a/lib/configs/typescript.js
+++ b/lib/configs/typescript.js
@@ -4,6 +4,7 @@ module.exports = {
   plugins: ['@typescript-eslint', 'github'],
   rules: {
     camelcase: 'off',
+    'no-unused-vars': 'off',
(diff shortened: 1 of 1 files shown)


keithamus:
> @typescript-eslint already includes a no-unused-vars rule so we use that one instead.

For posterity - we extend from [`@typescript-eslint/recommended` which includes the `@typescript/no-unused-vars` rule.](https://github.com/typescript-eslint/typescript-eslint/blob/1f0ff41aa1bc3b7c5330b2f5fe22e24bf578a6b2/packages/eslint-plugin/src/configs/recommended.json#L25)

Issue #89, opened 2020-03-02
bin/eslint-unused-modules does not check .ts files


muan:
https://github.com/github/eslint-plugin-github/blob/master/bin/eslint-unused-modules.js

All the exports in our new TS files are currently unchecked. 😬

koddsson:
I think this is because eslint defaults to files with a `.js` extension. Since our script takes in file globs we could just pass globs that include the `.ts` extension as a `bin/eslint-unused-modules src/**/*.ts`.

I looked into this a bit though and was hoping we could drop our custom solution for something built into TypeScript. I didn't find anything build into TypeScript but I did find [this gist](https://gist.github.com/dsherret/0bae87310ce24866ae22425af80a9864) that seems to yield correct results. I also timed the gist against our solution and the gist clocks in at around 13 seconds whereas our solution takes 50 seconds.

koddsson:
I think https://github.com/github/github/pull/140799 is a good indicator that we might want this check still.

PR #105, opened 2020-04-14
Add script that outputs exported functions that aren't imported

koddsson:
Replaces the script we removed after a migration to TypeScript. Modified from https://gist.github.com/dsherret/0bae87310ce24866ae22425af80a9864.

Fixes https://github.com/github/eslint-plugin-github/issues/89


koddsson:
Hmm, maybe this should just live in projects that want to use it rather in this eslint config 🤔

keithamus:
Yes let's move this to its own module.

PR #106, opened 2020-05-11
Turn off `@typescript-eslint/no-empty-function'` rule.

koddsson:
We seem to overwrite this rule in `github/github` all the time with no clear path of removing the disable comments so I figure this rule isn't doing anything and can safely be turned off.

### DIFF
--- a/lib/configs/typescript.js
+++ b/lib/configs/typescript.js
@@ -4,12 +4,12 @@ module.exports = {
   plugins: ['@typescript-eslint', 'github'],
   rules: {
     camelcase: 'off',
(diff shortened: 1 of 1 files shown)


koddsson:
This is the last bit for 4.0.0.

dgraham:
The function can contain a comment explaining why it's empty, right? What is an example empty function that we want to allow?

koddsson:
> The function can contain a comment explaining why it's empty, right?

Oh I thought the rule required some code in it and having comment in it wouldn't fix the violations. With that discovery, I don't think this is needed.

PR #108, opened 2020-05-26
Don't extends configs in the browser and internal configs

koddsson:
It's not apparent to consumers of this plugin that when they extend the `browser` or `internal` configs they are extending the `recommended` config as well.

By not extending them in those configs we make the consumers denote each config they want to extend and give them the freedom to compose those configs together however they want.

This is a breaking change for consumers that have been relying on getting the recommended rules through the other configs.


koddsson:
For posterity: Me and & @muan talked in slack and decided to release this as a bug fix to v4 rather than releaseing v5.

PR #110, opened 2020-07-14
Add event listener lints

keithamus:
### What?

This adds 3 new rules to our `config/browser` eslint config:

  - `github/no-useless-passive`
  - `github/require-passive-events`
  - `github/prefer-observers`

To read more about these rules, see the documentation md files in `docs/`

### Why?

These represent our ideal best practices of how to author code for the GitHub frontend, and as such we'd like to enforce these rules with static analysis.


lib/configs/browser.js
```diff
@@ -10,6 +10,9 @@ module.exports = {
     'github/no-blur': 'error',
     'github/no-dataset': 'error',
     'github/no-innerText': 'error',
-    'github/unescaped-html-literal': 'error'
+    'github/unescaped-html-literal': 'error',
+    'github/no-useless-passive': 'error',
+    'github/no-non-passive-high-frequency-events': 'error',
+    'github/prefer-observers-over-events': 'error'
```
keithamus, line 16: Do we consider this to be a breaking change? Do we want to release this as semver major?

keithamus: To answer my own question, precedent exists for this to be semver minor, see https://github.com/github/eslint-plugin-github/pull/80 which was released as `3.3.0`

lib/configs/browser.js
```diff
@@ -10,6 +10,9 @@ module.exports = {
     'github/no-blur': 'error',
     'github/no-dataset': 'error',
     'github/no-innerText': 'error',
-    'github/unescaped-html-literal': 'error'
+    'github/unescaped-html-literal': 'error',
+    'github/no-useless-passive': 'error',
+    'github/no-non-passive-high-frequency-events': 'error',
```
dgraham, line 15: Can we replace the double negative with a positive assertion: require-passive-events?

lib/configs/browser.js
dgraham, line 16: We can tighten this up to something like `prefer-observers`. Lint rules typically name the alternatives in the error message rather than the rule name.

PR #118, opened 2021-03-30
fix: stop svg-element-attributes v2.0.0 from installing

andykenward:
When using the package in a new project `svg-element-attributes` `v2.0.0` is installed causing eslint to error with the below error. This is because `svg-element-attributes` `v2.0.0` is [ESM only](https://github.com/wooorm/svg-element-attributes/tree/ff6f55b696b754d8f3ea22891e86ecb40247c702#install) and must use `import` instead of `require` when using.

Doing a `npm i -D svg-element-attributes@^1.3.1` to the new project solves the issue.

I suggest changing the semver version of svg-element-attributes from `>=1.3.1` to `^1.3.1`. To avoid users of this package from having this issue.

`node v14.16.0`
`npm 6.14.12`

```sh
...node_modules/svg-element-attributes/index.js
require() of ES modules is not supported.
require() of example-repo/node_modules/svg-element-attributes/index.js from example/node_modules/eslint-plugin-github/lib/rules/get-attribute.js is an ES module file as it is a .js file whose nearest parent package.json contains “type”: “module” which defines all .js files in that package scope as ES modules.
Instead rename index.js to end in .cjs, change the requiring code to use import(), or remove “type”: “module” from example-repo/node_modules/svg-element-attributes/package.json.

    at Object.Module._extensions..js (internal/modules/cjs/loader.js:1216:13)
    at Module.load (internal/modules/cjs/loader.js:1039:32)
    at Module._load (internal/modules/cjs/loader.js:932:14)
    at Function.f._load (electron/js2c/asar_bundle.js:5:12738)
    at Module.require (internal/modules/cjs/loader.js:1079:19)
    at require (internal/modules/cjs/helpers.js:72:18)
    at Object.<anonymous> (example-repo/node_modules/eslint-plugin-github/lib/rules/get-attribute.js:1:30)
    at Module._compile (internal/modules/cjs/loader.js:1199:30)
    at Object.Module._extensions..js (internal/modules/cjs/loader.js:1220:10)
    at Module.load (internal/modules/cjs/loader.js:1039:32)

```

Signed-off-by: Andy Kenward <4893048+andykenward@users.noreply.github.com>

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,4 +1,0 @@
-language: node_js
-node_js:
-  - "10"
-  - "12"
--- a/README.md
+++ b/README.md
@@ -1,5 +1,7 @@
 # eslint-plugin-github
 
+[![Node CI](https://github.com/github/eslint-plugin-github/actions/workflows/nodejs.yml/badge.svg)](https://github.com/github/eslint-plugin-github/actions/workflows/nodejs.yml)
+
 ## Installation
 
 ```sh
--- a/package-lock.json
+++ b/package-lock.json
@@ -3048,9 +3048,9 @@
       }
     },
     "node_modules/y18n": {
-      "version": "4.0.0",
-      "resolved": "https://registry.npmjs.org/y18n/-/y18n-4.0.0.tgz",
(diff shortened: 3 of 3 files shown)


koddsson (approved):
Looks good!

PR #122, opened 2021-03-31
Add info into the `meta` property of rules

koddsson:
It's good to have this info filled out. I was also hoping we could add https://github.com/not-an-aardvark/eslint-plugin-eslint-plugin to lint these rules and their tests; this change is in preparation for that.

## References
https://eslint.org/docs/developer-guide/working-with-rules#rule-basics

### DIFF
--- a/lib/rules/array-foreach.js
+++ b/lib/rules/array-foreach.js
@@ -3,7 +3,7 @@ module.exports = {
     type: 'suggestion',
     docs: {
       description: 'enforce `for..of` loops over `Array.forEach`',
-      url: 'https://github.com/github/eslint-plugin-github/blob/main/docs/rules/array-foreach.md'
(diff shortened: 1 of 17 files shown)


lib/rules/array-foreach.js
```diff
@@ -1,6 +1,10 @@
 module.exports = {
   meta: {
-    docs: {},
+    type: 'suggestion',
+    docs: {
+      description: 'enforce `for..of` loops over `Array.forEach`',
+      url: 'https://github.com/github/eslint-plugin-github/blob/main/docs/rules/array-foreach.md'
```
koddsson, line 6: These should be automatically generated.

keithamus (approved):
Looks like a great cleanup!

PR #128, opened 2021-07-15
Add more rules

koddsson:
We've slowly added more and more rules to the app config in `github/github`. It's good to have all the rules in one place to have a better overview of them and share the config with other applications.

This PR takes many rules from the `github/github` project and adds them into the recommended and browser configs.


bscofield (approved):
🎉

PR #127, opened 2021-07-15
Fix no shadow rule for typescript

koddsson:
The `no-shadow` rule from ESLint Core will report on lines that don't violate the rule when run on TypeScript files. We should use the `typescript-eslint` rule for `no-shadow` rather than the rule in ESLint Core.

I also threw in a quick `npm audit fix` while I was at it.

## References
https://github.com/typescript-eslint/typescript-eslint/blob/master/docs/getting-started/linting/FAQ.md#i-am-using-a-rule-from-eslint-core-and-it-doesnt-work-correctly-with-typescript-code
Fixes #126

PR #129, opened 2021-07-16
fixes typescript named/imports for github/typescript

stephenotalora:
The `import/named` rule from ESLint Core will report on lines that don't violate the rule when run on TypeScript files. We should probably do something similar to our `no-shadow fix`

## References

https://github.com/github/eslint-plugin-github/pull/127

### DIFF
--- a/lib/configs/typescript.js
+++ b/lib/configs/typescript.js
@@ -5,6 +5,8 @@ module.exports = {
   rules: {
     camelcase: 'off',
     'no-unused-vars': 'off',
+    'import/named': 'off',
+    '@typescript-eslint/import/named': ['error'],
     '@typescript-eslint/interface-name-prefix': 'off',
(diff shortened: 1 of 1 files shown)


koddsson:
Is there a rule called `import/named` in `@typescript-eslint`? I couldn't find it [in the repository](https://github.com/typescript-eslint/typescript-eslint/tree/master/packages/eslint-plugin/src/rules).

stephenotalora:
> Is there a rule called `import/named` in `@typescript-eslint`? I couldn't find it [in the repository](https://github.com/typescript-eslint/typescript-eslint/tree/master/packages/eslint-plugin/src/rules).

good catch, this only silences the error. it looks like we needed another plugin of the parser to properly work with named imports. Extending from `'plugin:import/typescript'` fixes this problem. 

Not PR needed here closing in favor of https://github.com/benmosher/eslint-plugin-import/blob/master/README.md#typescript

PR #132, opened 2021-08-09
Add no-only-tests JS linter rule to detect test.only and suite.only

pablonete:
This warns on leftovers like `suite.only` or `test.only` that would prevent all JS tests from running in CI.

Adding a dependency on this plugin: https://www.npmjs.com/package/eslint-plugin-no-only-tests
It meets the "LAWSUIT" external library criteria:
- **L**icensing: MIT licensed
- **A**ctively-developed: Latest commit 1 month ago, on June 17th
- **W**idely-used: 107,289 weekly downloads
- **S**upply chain safety: only dependency is "eslint"
- **U**nique functionality: Yes
- **I**nstall time impact + bundle size impact: Dev-time linter, no production impact.
- **T**ypeScript support: this doesn't apply to an eslint plugin.

By default, the library doesn't check `suite`, so I had to change the default value of `block` by adding it to the default values:
```
"block": ["describe", "it", "context", "test", "tape", "fixture", "serial", "suite"]
```

Tested with `npm link` successfully:
image

> Note: I wasn't able to get `npm link` to work in Codespaces, I had to use my laptop. I got:
> npm was not found in PATH outside of this project
> I suspect it's due to `npm` not being installed in the dotcom Codespace, and even using the `bin/npm` in the $PATH didn't help.

### DIFF
--- a/.eslintrc.js
+++ b/.eslintrc.js
@@ -8,13 +8,19 @@ module.exports = {
     node: true
   },
   extends: [require.resolve('./lib/configs/recommended'), 'plugin:eslint-plugin/all'],
-  plugins: ['eslint-plugin'],
+  plugins: ['eslint-plugin', 'no-only-tests'],
   rules: {
     'import/no-commonjs': 'off',
     'filenames/match-regex': 'off',
     'i18n-text/no-en': 'off',
     'eslint-plugin/prefer-placeholders': 'off',
     'eslint-plugin/test-case-shorthand-strings': 'off',
-    'eslint-plugin/require-meta-docs-url': 'off'
+    'eslint-plugin/require-meta-docs-url': 'off',
+    'no-only-tests/no-only-tests': [
+      'error',
+      {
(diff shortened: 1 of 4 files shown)


.eslintrc.js
```diff
@@ -8,13 +8,19 @@ module.exports = {
     node: true
   },
   extends: [require.resolve('./lib/configs/recommended'), 'plugin:eslint-plugin/all'],
-  plugins: ['eslint-plugin'],
+  plugins: ['eslint-plugin', 'no-only-tests'],
```
koddsson, line 11: This change adds the rule to the ESLint config of this project. We should add it to some configs we provide as well. Can you make sure to include the rule in `lib/configs/recommended.js` as part of this pull request?

pablonete: Of course :facepalm:
Indeed I've removed it from this project ESLint config, because tests here use `ruleTester.run` which won't get advantage of the new no-only-tests rule.

koddsson (approved):
This is great! Thank you ❤️

PR #137, opened 2021-09-09
Add no innerHTML rule

manuelpuyol:
Adds a rule that disallows the usage of `innerHTML` with some kind of content.
The only allowed use is `innerHTML = ''
`

```js
// bad
function setContent(element, content) {
  element.innerHTML = content
}
// good
function clearContent(element) {
  element.innerHTML = ''
}
```

### DIFF
--- a/lib/rules/no-inner-html.js
+++ b/lib/rules/no-inner-html.js
@@ -1,0 +1,33 @@
+module.exports = {
+  meta: {
+    type: 'problem',
+    docs: {
+      description: 'disallow `Element.prototype.innerHTML``',
+      url: require('../url')(module)
+    },
+    schema: []
+  },
+
+  create(context) {
+    return {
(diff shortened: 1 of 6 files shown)


lib/rules/no-inner-html.js
```diff
@@ -0,0 +1,33 @@
...
+
+          if (leftNode.property && leftNode.property.name === 'innerHTML') {
+            if (rightNode.type === 'Literal' && rightNode.value === '') {
+              return
+            }
+
+            context.report({
+              node,
+              message:
+                'Using innerHTML poses a potential security risk and should not be used other than cleaning content.'
```
dgreif, line 26: Maybe `clearing` instead of `cleaning`?

dgreif (approved):
Looks like a great rule to add!

PR #138, opened 2021-09-10
Suggest using textContent instead of innerHTML

manuelpuyol:
We were suggesting using `innerHTML` to clear content, but that's the same as `textContent = ''`.
This PR will block **all** uses of `innerHTML` and autocorrect to use `textContent` instead

### DIFF
--- a/docs/rules/no-innter-html.md
+++ b/docs/rules/no-innter-html.md
@@ -1,6 +1,6 @@
 # No `innerHTML`
 
-Using `innerHTML` poses a potential security risk. Prefer using `textContent` so set text to an element.
+Using `innerHTML` poses a potential security risk. Prefer using `textContent` to set text to an element.
 
 ```js
 // bad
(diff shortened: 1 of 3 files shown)


dgreif (approved):
Nice, I like the change!

docs/rules/no-innter-html.md
```diff
@@ -1,6 +1,6 @@
 # No `innerHTML`
 
-Using `innerHTML` poses a potential security risk. It should only be used when clearing content.
+Using `innerHTML` poses a potential security risk. Prefer using `textContent` so set text to an element.
```
JoseInTheArena, line 3: Did you mean here "_Prefer using `textContent` **to** set text to an element_"?

JoseInTheArena (commented):
Thanks for putting this together! 🙇🏻‍♂️ I just have a small question about the copy on the docs.

docs/rules/no-innter-html.md
manuelpuyol:
```suggestion
Using `innerHTML` poses a potential security risk. Prefer using `textContent` to set text to an element.
```

manuelpuyol:
per @koddsson request, I removed autocorrection since it was confusing

PR #145, opened 2021-11-03
Fix the version of eslint-plugin-import

yuya-takeyama:
I don't know why but `package.json` and `package-lock.json` look strange.

In the latest release v4.3.3, `eslint-plugin-import` is set to `"^2.24.2"`:
https://github.com/github/eslint-plugin-github/blob/v4.3.3/package.json#L34
But the version specified in `pakage-lock.json` is `"2.25.2"`:
https://github.com/github/eslint-plugin-github/blob/v4.3.3/package-lock.json#L954

Also, dependency to `typescrpit` disappeared by running `npm i`.
As far as I can see, there is no package depends on `typescript`.
So I think this is correect.

image

I'm using npm v7.24.2.

---

I hope that this repo will use dependabot or renovate to reduce manual modifications for dependencies.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -15,7 +15,7 @@
         "eslint-plugin-eslint-comments": "^3.2.0",
         "eslint-plugin-filenames": "^1.3.2",
         "eslint-plugin-i18n-text": "^1.0.1",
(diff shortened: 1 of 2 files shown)


keithamus (approved):
:+1:

PR #166, opened 2021-12-08
chore: Minor typo cleanup

nschonni:
Took a quick pass with cSpell

### DIFF
--- a/docs/rules/array-foreach.md
+++ b/docs/rules/array-foreach.md
@@ -13,7 +13,7 @@ Here's a summary of why `forEach` is disallowed, and why we prefer `for...of` fo
 
 Typically developers will reach for a `forEach` when they want to iterate over a set of items. However not all "iterables" have access to Array methods. So a developer might convert their iterable to an Array by using `Array.from(iter).forEach()`. This code has introduced performance problems, where a `for...of` loop would be more performant.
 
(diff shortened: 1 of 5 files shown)


tests/require-passive-events.js
```diff
@@ -47,7 +47,7 @@ ruleTester.run('require-passive-events', rule, {
       ]
     },
     {
-      // Intentionally mispelled!
+      // Intentionally misspelled!
```
srt32, line 50: :)

docs/rules/array-foreach.md
```diff
@@ -58,7 +58,7 @@ for (const apple of apples) {
 }
 ```
 
-If `polishApple` needed to do some serial async work, then we'd need to refactor the iteration steps to accomodate for this async work, by `await`ing each call to `polishApple`. We cannot simply pass an `async` function to `forEach`, as it does not understand async functions, instead we'd have to turn the `forEach` into a `reduce` and combine that with a `Promise` returning function. For example:
+If `polishApple` needed to do some serial async work, then we'd need to refactor the iteration steps to accommodate for this async work, by `await`ing each call to `polishApple`. We cannot simply pass an `async` function to `forEach`, as it does not understand async functions, instead we'd have to turn the `forEach` into a `reduce` and combine that with a `Promise` returning function. For example:
 
 ```diff
 - apples.forEach(polishApple)
```
srt32, line 64: Could you explain this diff here?

srt32 (commented):
Great catches!

docs/rules/array-foreach.md
nschonni: The line you're commenting on is a code block with `diff` language tag, not something I changed

tests/require-passive-events.js
nschonni: Yeah, I didn't correct the intentional typo below :)

docs/rules/array-foreach.md
srt32: aha! Yes.

srt32 (approved):

PR #185, opened 2022-01-21
Allow files to be named _.config.js

keithamus:
We allow many files to be named `*.config.js`, e.g. `postcss.config.js`, `rollup.config.js`, `karma.config.cjs`. 

This should be allowed in the lint rule, without having to override it.


koddsson (approved):
Looks good!

PR #192, opened 2022-02-03
Prefer import extensions

koddsson:
A rule which enforces extensions on file imports. It's not perfect since it throws a `.js` extension on the import literal in the autofix but I'm unsure on the heuristics for our edge cases. I think in most cases we want imports to just have `.js` but this might change in the future, especially with [import assertions](https://github.com/tc39/proposal-import-assertions) on the horizon.


koddsson:
Thinking about this, I wonder if I should align this further to the imports assertion specifications to future proof this rule.

PR #223, opened 2022-03-24
Add rule to prohibit building script tags in the client

manuelpuyol:
We should never allow building `script` tags in our applications since they side-step some security measures.


lib/rules/no-script-tag-building.js
```diff
@@ -0,0 +1,30 @@
+module.exports = {
+  meta: {
+    type: 'suggestion',
+    docs: {
+      description: 'disallow marking a event handler as passive when it has no effect',
```
theinterned, line 5: does this need to be updated?

README.md
```diff
@@ -53,6 +53,7 @@ The available configs are:
 - [No Implicit Buggy Globals](./docs/rules/no-implicit-buggy-globals.md)
 - [No Inner HTML](./docs/rules/no-inner-html.md)
 - [No InnerText](./docs/rules/no-innerText.md)
+- [No Script Tag Building](./docs/rules/no-script-tag-building.md)
```
theinterned, line 56: 🚲 🏠 Def a bike shed so disregard at your discretion, but maybe this should be called **"No Dynamic Script Tags"**? I notice that this is in fact what you refer to them as in all the description text in this PR for example.

theinterned (approved):
❤️ Thank you for addressing my feedback!

lib/rules/no-dynamic-script-tag.js
```diff
@@ -0,0 +1,30 @@
+module.exports = {
+  meta: {
+    type: 'suggestion',
+    docs: {
+      description: 'disallow creating dynamic script tags',
+      url: require('../url')(module)
+    },
+    fixable: 'code',
```
koddsson, line 8: This isn't fixable.

```suggestion
```

tests/no-dynamic-script-tag.js
```diff
@@ -0,0 +1,38 @@
...
+      code: 'document.createElement("script")',
+      errors: [
+        {
+          message: "Don't create dynamic script tags, add them in the server template instead.",
+          type: 'Literal'
+        }
+      ]
+    },
+    {
+      code: 'document.createElement("span").type = "text/javascript"',
```
koddsson, line 29: This should be a `script` right? Spans don't have a `type`.

manuelpuyol: I just wanted to test that `.type = "text/javascript"` will fail. If I add `document.createElement("script")`, it will fail without checking the `type`

PR #282, opened 2022-07-19
Add accessibility rule and React config

khiga8:
Relates to: [github/accessibility#1403](https://github.com/github/accessibility/issues/1403) (GitHub staff only)

CC: @github/accessibility-reviewers too for any thoughts

## Motivation

We want to start creating and sharing custom accessibility rules for React projects. While there is a `jsx-plugin-a11y`, GitHub also has some opinionated stances when it comes to accessibility which we want to codify with a linter.

I noticed that GitHub already owns a couple open-source eslint repos so I decided to start adding accessibility rules in this repo rather than starting from scratch. I am thinking that non Primer-specific accessibility rules belong in this repo, while Primer-specific accessibility rules can go in [eslint-plugin-primer-react](https://github.com/primer/eslint-plugin-primer-react). GitHub staff should check out the following links for more context:

- [comment in issue 1403](https://github.com/github/accessibility/issues/1403#issuecomment-1188253148).
- [slack thread](https://github.slack.com/archives/CMZ4DC9BL/p1658175507494509)

## Changes

I am trying to set up a foundation for more accessibility rules to be added and shared between GitHub React projects.

### New accessibility rule

I am adding an accessibility rule that discourages use of generic link text in JSX elements. This is the equivalent of an existing ERB lint rule we have ([avoid-generic-link-text-counter.md](https://github.com/github/erblint-github/blob/main/docs/rules/accessibility/avoid-generic-link-text-counter.md)). We want to flag when a link element has an `aria-label` with generic link text, or when there is no `aria-label` set but it has a text content of a generic link text. 

This is the first time I am writing an eslint rule so please review carefully. 

### New config

Should I name this to something specific to JSX accessibility linting like `jsx-a11y` so the config is: `github/jsx-a11y`? 
Or is generic `react` fine?

## Note to reviewer 📣

I'm pretty new to eslint and eslint best practices so if any of this organization seems really funky, please let me know. 

Also if you think accessibility rulesets should be organized in a different way or should be in its own repo talk to me! I am in spike phase right now and very open to suggestions/iteration.


lib/index.js
```diff
@@ -9,6 +9,7 @@ module.exports = {
     'no-blur': require('./rules/no-blur'),
     'no-d-none': require('./rules/no-d-none'),
     'no-dataset': require('./rules/no-dataset'),
+    'no-generic-link-text': require('./rules/no-generic-link-text'),
```
khiga8, line 12: I want this to be prefixed so I can pattern match the code base specifically for accessibility rules. What should I do???

[Related Slack thread on prefixing eslint rules ](https://github.slack.com/archives/CMZ4DC9BL/p1658186503467939?thread_ts=1658175507.494509&cid=CMZ4DC9BL)(GitHub Staff only)

lib/configs/react.js
```diff
@@ -0,0 +1,10 @@
+module.exports = {
+  env: {
+    browser: true
+  },
+  plugins: ['github', 'jsx-a11y'],
+  extends: ['plugin:jsx-a11y/recommended'],
```
khiga8, line 6: I'm assuming that any config we want to recommend as part of "react" belongs here, so I added the `jsx-a11y` plugin. does this look right?

docs/rules/no-generic-link-text.md
```diff
@@ -0,0 +1,83 @@
+# No Generic Link Text
```
khiga8, line 1: This is pretty much copied from our `erblint-github` docs though the examples were changed to `.jsx` instead of `.erb`. 

The two behave nearly exactly the same so why not 🤷🏻‍♀️

package.json
```diff
@@ -33,9 +33,11 @@
     "eslint-plugin-filenames": "^1.3.2",
     "eslint-plugin-i18n-text": "^1.0.1",
     "eslint-plugin-import": "^2.25.2",
+    "eslint-plugin-jsx-a11y": "^6.6.0",
     "eslint-plugin-no-only-tests": "^2.6.0",
     "eslint-plugin-prettier": "^4.0.0",
     "eslint-rule-documentation": ">=1.0.0",
+    "jsx-ast-utils": "^3.3.2",
```
khiga8, line 40: This provides utility methods that make parsing JSX elements much more pleasant. I found it through `jsx-a11y-plugin`. 

[Repo for jsx-ast-utils](https://github.com/jsx-eslint/jsx-ast-utils)

lib/index.js
khiga8: I changed the name to `a11y-`

lib/configs/react.js
```diff
@@ -0,0 +1,10 @@
+module.exports = {
```
khiga8, line 1: Maybe I should distinguish this config as being accessibility rules for react....like `github/jsx-a11y`. 

thoughts? @github/accessibility-reviewers

lib/index.js
kendallgassner: I like the a11y!

package.json
kendallgassner: do u also need to use the typescript ast?

lib/configs/react.js
kendallgassner: I think for future rules github/react is kinda nice?

package.json
khiga8: It seems yes, as long as the `.eslintrc.json` wherever the linters get run is configured with the typescript parser:

```
 "parser": "@typescript-eslint/parser",
```

I tested this branch out in dotcom by running:

```
npm i git://github.com/github/eslint-plugin-github.git#kh-add-a11y --save
```

(^today I learned how to install a package from a GitHub branch!!!)


I see the lint error show up as expected in my editor!

Screenshot of my VSCode editor in dotcom which flags an eslint violation from this rule

lib/configs/react.js
khiga8: Yeah I'm not sure either... I'm down to keep this for now and revisit later!

package.json
kendallgassner: 🕺 ! YAY thanks for checking

khiga8:
Hi 👋

This is ready for review. @github/web-systems-reviewers. I plan to follow up on this by creating a config where we can map Primer components to a tag.

lib/rules/a11y-no-generic-link-text.js
```diff
@@ -0,0 +1,68 @@
+const {elementType, getProp, getPropValue} = require('jsx-ast-utils')
+
+const bannedLinkText = ['read more', 'here', 'click here', 'learn more', 'more', 'here']
```
jbrown1618, line 3: nit: `'here'` is in this list twice.

lib/rules/a11y-no-generic-link-text.js
```diff
@@ -0,0 +1,68 @@
...
+      description: 'disallow generic link text',
+      url: require('../url')(module)
+    },
+    schema: []
+  },
+
+  create(context) {
+    return {
+      JSXOpeningElement: node => {
+        if (elementType(node) !== 'a') return
```
jbrown1618, line 26: In general we will probably use the `Link` component from `@primer/react` more often than raw anchor tags, and this check would miss those.

I don't think we can catch all of the possible cases just using a linter, but it might be more thorough to apply this check to any element with an `href` prop. That will at least catch the cases where we have components that are thin wrappers around anchor tags.

khiga8: I have a follow up PR to address this. I'd love your feedback there -- https://github.com/github/eslint-plugin-github/pull/283.

jfuchs (approved):

PR #284, opened 2022-07-21
Update README with component mapping setting

khiga8:
## Context

In https://github.com/github/eslint-plugin-github/pull/283 we introduced a config that allows custom components to be mapped to an HTML element for our JSX rules.

## Changes

This adds a README that explains how the component mapping works. I marked this feature as experimental because it is new and subject to change.

This PR also updates the `parserOptions` for the `react` config. Let me know if this looks wrong though.

## Note

I plan to cut a new release following this PR.

### DIFF
--- a/README.md
+++ b/README.md
@@ -41,6 +41,35 @@ The available configs are:
 - `typescript`
   - Useful rules when writing TypeScript.
 
+### Component mapping (Experimental)
+
+_Note: This is experimental and subject to change._
+
+The `react` config includes rules which target specific HTML elements. You may provide a mapping of custom components to an HTML element in your `eslintrc` configuration to increase linter coverage.
+
+For each component, you may specify a `default` and/or `props`. `default` may make sense if there's a 1:1 mapping between a component and an HTML element. However, if the HTML output of a component is dependent on a prop value, you can provide a mapping using the `props` key. To minimize conflicts and complexity, this currently only supports the mapping of a single prop type.
+
+```json
+{
+  "settings": {
+    "github": {
+      "components": {
+        "Box": { "default": "p" },
+        "Link": { "props": {"as": { "undefined": "a", "a": "a", "button": "button"}}},
+      }
+    }
+  }
+}
+```
+
+This config will be interpreted in the following way:
+
+- All `<Box>` elements will be treated as a `p` element type.
+- `<Link>` without a defined `as` prop will be treated as a `a`.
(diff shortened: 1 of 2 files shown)


lib/configs/react.js
```diff
@@ -1,6 +1,9 @@
 module.exports = {
-  env: {
-    browser: true
+  parserOptions: {
+    sourceType: 'module',
+    ecmaFeatures: {
+      jsx: true
+    }
```
khiga8, line 6: I _believe_ this is the configuration we want for React linting based on [language options guide](https://eslint.org/docs/latest/user-guide/configuring/language-options) and the config in [primer-react/recommended](https://github.com/primer/eslint-plugin-primer-react/blob/main/src/configs/recommended.js).

README.md
```diff
@@ -41,6 +41,35 @@ The available configs are:
...
+}
+```
+
+This config will be interpreted in the following way:
+
+- All `<Box>` elements will be treated as a `p` element type.
+- `<Link>` without a defined `as` prop will be treated as a `a`.
+- `<Link as='a'>` will treated as an `a` element type.
+- `<Link as='button'>` will be treated as a `button` element type.
+- `<Link as='summary'>` will be treated as the original `Link` element type because there is no matching mapping for `as='summary'`.
```
owenniblock, line 71: Nice, do you mean here that it will be treated as an `a`? If so, I suggest we change this slightly to make it explicit:

```suggestion
- `<Link as='summary'>` will be treated as the original `Link` element type because there is no matching mapping for `as='summary'` resulting in the element being treated as an `a`
```

Also, very minor but I think it should be "as an `a`" rather than "as a".

owenniblock (approved):
Great docs (especially the note on is the `as` isn't defined. I can't comment on the configuration change unfortunately, is there an easy way to test it before we merge?

lindseywild (approved):
Thanks for adding these docs! ✨

khiga8:
@owenniblock I recently just learned that you can install npm packages from a github branch! I ran the following on dotcom:

```
npm i git://github.com/github/eslint-plugin-github.git#kh-add-docs-for-component-config  --save
```

The lint ran fine and caught some issues. But also I didn't run into issues with the previous config so I'm not really sure what I'm supposed to expect. This looks more right though...

khiga8:
I guess I'll test this in `memex` too! If it looks good I shall merge this.

PR #343, opened 2022-11-17
allow `.server.[ext]` file extension

theinterned:
This change updates the recommended eslint config to match the `.server.[ext]` file extension introduced in https://github.com/github/github/pull/247398 ... although the actual extension is under discussion and may change.


theinterned:
I am thinking I want to generalize this PR to generically allow `*.*.*$` — I ie `file-name.prefix.ext` as we are seeing more examples of this naming scheme: `*.test.ts`, `*.server.tsx`, `*.config.json`

dgreif:
While we may indeed need to expand the list, I'd rather do so as needed to avoid weird/unexpected file types showing up

dgreif (approved):

PR #354, opened 2022-12-05
Update events handled by `no-useless-passive` and `require-passive-events`

boris-petrov:
Check:
https://github.com/ignasdamunskis/passive-events-support/blob/65879a4a57a01763c9a8a07a762f60a2ac268f53/src/utils.js#L24

I'm not sure which list is "more correct"... but I guess more is better, right? :smile:


keithamus (approved):
LGTM :smile:

keithamus (approved):
Still LGTM :+1:

dgreif (approved):

PR #369, opened 2022-12-28
drop node 12 / add node 18 to test matrix

theinterned:
[I noticed in this PR](https://github.com/github/eslint-plugin-github/pull/368#issuecomment-1366965933) that tests are failing in node 12 but runnign successfully in node 14, 16, and 18.

But I think we should just change the test node matrix to 14, 16, and 18 seeing as [12 is End-of-Life while 18 is LTS](https://github.com/nodejs/release#release-schedule).

### DIFF
--- a/.github/workflows/nodejs.yml
+++ b/.github/workflows/nodejs.yml
@@ -12,7 +12,7 @@ jobs:
 
     strategy:
       matrix:
-        node-version: [12.x, 14.x, 16.x]
+        node-version: [14.x, 16.x, 18.x]
 
     steps:
       - uses: actions/checkout@v3


.github/workflows/nodejs.yml
```diff
@@ -12,7 +12,7 @@ jobs:
 
     strategy:
       matrix:
-        node-version: [12.x, 14.x, 16.x]
+        node-version: [14.x, 16.x, 18.x]
```
nschonni, line 15:
```suggestion
        node-version:
          - 14
          - 16
          - 18
          - 19
```

theinterned: I like it!

theinterned: In fact, why not make the node version dynamic? https://github.com/github/eslint-plugin-github/actions/runs/3802048520

nschonni: Up to you, but I think it's easier to understand the actual versions, and sync it with an `engines` value in https://github.com/github/eslint-plugin-github/blob/main/package.json

.github/workflows/nodejs.yml
```diff
@@ -12,7 +12,11 @@ jobs:
 
     strategy:
       matrix:
-        node-version: [12.x, 14.x, 16.x]
+        node-version:
```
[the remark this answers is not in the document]
smockle: Because the matrix includes `latest`, when `lts/*` begins pointing to a new version, no failures should occur, because that new version would have already been tested, back when it was `latest`. So non-fixed (_viz._ `lts/-2`, `lts/-1`, `lts/*`) values ought to be safe.

That said, when `latest` itself begins pointing to a new version, failures might occur. But if we remove `latest`, then `lts/*` loses the safety described above.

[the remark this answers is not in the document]
smockle: > when `latest` itself begins pointing to a new version, failures might occur

Here’s one workaround: For `latest` (and only `latest`), if any step of the job fails ([`failure()`](https://docs.github.com/en/actions/learn-github-actions/expressions#failure)), [open an issue](https://github.com/github/accessibility-scorecard/blob/main/.github/actions/report-failure/action.yml), then `exit 0` so the overall workflow run is still green.

nschonni:
You seem to be going to a lot of effort just to avoid using the version numbers in the matrix. Those changes happen infrequentely and are also a good time to evaluated if major versions of things like ESLint need to change. They usually have a major bump when version support changes

.github/workflows/nodejs.yml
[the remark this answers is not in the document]
theinterned: Okay! cool. I got this working!

Here's a sample failed run: https://github.com/github/eslint-plugin-github/actions/runs/3884555154, and the issue it created: https://github.com/github/eslint-plugin-github/issues/378

To test this I created a failing test https://github.com/github/eslint-plugin-github/pull/369/commits/5077f3006ce588328af93f049cae1ad49b781220 which I will now delete.

nschonni:
I'd suggest taking a look at what some of the community supported extensions do like https://github.com/eslint-community/eslint-plugin-promise/blob/main/.github/workflows/CI.yml
The CI matches the supported versions of ESLint, including some older ones that wont be dropped till a semver major release of ESLint. They also have cross version testing of the last major version of ESLint.
Since this is an internal plug-in, you've got more flexibility on choosing what you want to support though.

theinterned:
> You seem to be going to a lot of effort just to avoid using the version numbers in the matrix

Yeah maybe but it was a fun to work on little challenge and I learned some stuff 🤷 

> Those changes happen infrequentely and are also a good time to evaluated if major versions of things like ESLint need to change. They usually have a major bump when version support changes

Note that `current` doesn't fail CI — but this kind of signal could alert us that there' work to do to support the next version.

nschonni:
> Note that `current` doesn't fail CI — but this kind of signal could alert us that there's work to do to support the next version.

Yeah, but that support likely needs to come from ESLint first. Their matrix is also explicit, and is used to test their supported versions https://github.com/eslint/eslint/blob/2952d6ed95811ce0971b6855d66fb7a9767a7b72/package.json#L168-L170

nschonni:
Anyway, I'm going to unsubscribe from this as I think we're bikeshedding at this point, and it's not worth either of our time

.github/workflows/nodejs.yml
```diff
@@ -23,5 +27,21 @@ jobs:
...
       - name: Install
         run: npm ci
+      - name: Node version
+        id: version
+        run: echo "node-version=$(node -v)" >> $GITHUB_OUTPUT
       - name: Test
+        id: test
         run: npm test
+        continue-on-error: ${{ matrix.node-version == 'current' }}
+      - name: Create Issue on Fail
```
jfuchs, line 37: Is this going to create a new issue on every failure?

jfuchs (approved):
Question not blocking.

.github/workflows/nodejs.yml
theinterned: Yes it will. Can you think of a way to improve that?

theinterned:
Maybe this is all too much. Dropping this dynamic approach in favour of https://github.com/github/eslint-plugin-github/pull/380

It was still a fun little project.

PR #398, opened 2023-02-08
Automate docs with eslint-doc-generator

bmish:
[eslint-doc-generator](https://github.com/bmish/eslint-doc-generator) is a CLI tool I built for automating the generation of the README rules list table and rule doc title/notices for ESLint plugins. It follows common documentation conventions from this and other top ESLint plugins and will help us standardize documentation across ESLint plugins (and generally improve the usability of custom rules through better documentation and streamline the process of adding new rules). It has 100% test coverage and is [used](https://github.com/bmish/eslint-doc-generator#users) in many of the top ESLint plugins already.

### DIFF
--- a/docs/rules/a11y-no-generic-link-text.md
+++ b/docs/rules/a11y-no-generic-link-text.md
@@ -1,6 +1,8 @@
-# No Generic Link Text [Deprecated]
+# Disallow generic link text (`github/a11y-no-generic-link-text`)
 
-_This rule has been deprecated in favor of `anchor-ambiguous-text` in `eslint-plugin-jsx-a11y`._
+❌ This rule is deprecated. It was replaced by `jsx-a11y/anchor-ambiguous-text`.
+
+<!-- end auto-generated rule header -->
 
 ## Rule Details
 
--- a/.eslint-doc-generatorrc.js
+++ b/.eslint-doc-generatorrc.js
@@ -1,0 +1,9 @@
+/** @type {import('eslint-doc-generator').GenerateOptions} */
+module.exports = {
+  configEmoji: [
+    ['browser', '🔍'],
+    ['internal', '🔐'],
+    ['react', '⚛️']
+  ],
+  ruleDocSectionInclude: ['Rule Details', 'Version'],
+};
--- a/README.md
+++ b/README.md
@@ -3,7 +3,7 @@
 ## Installation
 
 ```sh
-$ npm install --save-dev eslint eslint-plugin-github
+npm install --save-dev eslint eslint-plugin-github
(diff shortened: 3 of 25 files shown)


docs/rules/a11y-no-generic-link-text.md
```diff
@@ -1,6 +1,8 @@
-# No Generic Link Text [Deprecated]
+# Disallow generic link text (`github/a11y-no-generic-link-text`)
```
theinterned, line 1: Are these title updates auto-generated by the tool? If so what are they based on?

docs/rules/a11y-no-generic-link-text.md
```diff
@@ -1,6 +1,8 @@
...
 
-_This rule has been deprecated in favor of `anchor-ambiguous-text` in `eslint-plugin-jsx-a11y`._
+❌ This rule is deprecated. It was replaced by `jsx-a11y/anchor-ambiguous-text`.
```
theinterned, line 3: what about this part? is it auto-generated? If so what is it based on?

theinterned:
Thank you for this contribution. In general this looks like a nice change. I just a have a few questions about what it means.

I guess the other question I have is: what maintenance process will this change? Should we update any docs to reflect those changes?

docs/rules/a11y-no-generic-link-text.md
bmish: Yep, based on the rule name and `meta.docs.description`, customizable here:

https://github.com/bmish/eslint-doc-generator#--rule-doc-title-format

docs/rules/a11y-no-generic-link-text.md
bmish: Yeah, based on `meta.deprecated` / `meta.replacedBy`.

bmish:
The only process change really is described here: 

https://github.com/bmish/eslint-doc-generator#usage

> Run the script from package.json to start out or any time you add a rule or update rule metadata in your plugin:
>
> ```
> npm run update:eslint-docs
> ```

Could be mentioned in `CONTRIBUTING.md` but wasn't sure if that was necessary. There's also a CI check that will fail and provide this instruction when the command needs to be run.

theinterned:
> There's also a CI check that will fail and provide this instruction when the command needs to be run

Ah this one https://github.com/github/eslint-plugin-github/pull/398/files#diff-7ae45ad102eab3b6d7e7896acd08c427a9b25b346470d7bc6507b6481575d519R15 ... I guess that covers my concern

theinterned (approved):

PR #418, opened 2023-03-23
New rule to flag invalid aria-label format

khiga8:
This new lint rule flags when `aria-label` content is not well-formatted. Specifically, it flags when a JSX opening tag has an `aria-label` content that is a `Literal`, and starts with downcase.

`aria-label` content should be formatted in the same way as one would visual text, but there seems to be a misconception that it doesn't have to be. As a result, we're seeing `aria-label` that looks like an ID, or is downcased. 

Staff-only: For fun, here's a list of [flagged `aria-label`](https://github.com/github/github/blob/dca80bff964a81502fc3a12efd686dc75f2ec2d1/flagged-aria-label-text.txt) in ERB using this logic. 

Comment from @jscholes:

> I don't know where the impression comes from that aria-label shouldn't be treated exactly like visual text in terms of formatting and such, but it should be.

Relates to: 
- https://github.com/github/accessibility/issues/3176
- [Staff-only Slack convo](https://github.slack.com/archives/C0FSWLQ0Y/p1679420729511839?thread_ts=1679395197.329979&cid=C0FSWLQ0Y)

I plan to follow-up with a similar rule in `erblint-github`.


README.md
```diff
@@ -82,27 +82,28 @@ This config will be interpreted in the following way:
...
-| [prefer-observers](docs/rules/prefer-observers.md)                   | disallow poorly performing event listeners                                                                               | 🔍 |    |    |
-| [require-passive-events](docs/rules/require-passive-events.md)       | enforce marking high frequency event handlers as passive                                                                 | 🔍 |    |    |
-| [role-supports-aria-props](docs/rules/role-supports-aria-props.md)   | Enforce that elements with explicit or implicit roles defined contain only `aria-*` properties supported by that `role`. | ⚛️ |    |    |
-| [unescaped-html-literal](docs/rules/unescaped-html-literal.md)       | disallow unescaped HTML literals                                                                                         | 🔍 |    |    |
+| Name                                                                     | Description                                                                                                              | 💼 | 🔧 | ❌  |
```
khiga8, line 85: this auto-generator is so cool!

khiga8:
I _think_ we can assume that there's no valid `aria-label` that is downcased that's hard-coded.

 If `aria-label` is populated based on something like a users content like a filename, that would be dynamically generated and would not be flagged by static analysis.

tests/a11y-aria-label-is-well-formatted.js
```diff
@@ -0,0 +1,29 @@
...
+    ecmaVersion: 'latest',
+    sourceType: 'module',
+    ecmaFeatures: {
+      jsx: true,
+    },
+  },
+})
+
+const errorMessage =
+  '[aria-label] text should be formatted the same as you would visual text. Use sentence case and make sure you are not using hyphens.'
```
kendallgassner, line 15: What about words with hyphens? Example: `six-pack`/

khiga8: Does this answer your concern: https://github.com/github/eslint-plugin-github/pull/418#issuecomment-1481685661? 

I think it's highly unlike we'll run into a valid, hard-coded `aria-label`. If we do run into one, we can consider adding a way to configure an exception list. Do you think this rule should ship with that?

khiga8: Theoretically, this rule would flag `six-pack` but it wouldn't flag `Six-pack` 😂.

kendallgassner: Hmmm, what if we also only flag these aria-labels if they don't have a single space?... Technically `six-pack` would still get flagged but it's a more safe check?

kendallgassner: `SIX-PACK`

khiga8: > what if we also only flag these aria-labels if they don't have a single space?...

I think we'd still want to flag something like:

```
show
hide
close
```

(these are examples I'm pulling from real aria-label we have right now in our codebase)

kendallgassner (approved):
approving -- we worked through the six-pack idea on slack

dgreif (approved):

PR #442, opened 2023-05-31
Update react.js to turn `jsx-a11y/no-autofocus` off

khiga8:
This is a follow-up to [Staff-only issue](https://github.com/github/accessibility/issues/2927).

There are legitimate uses of `autofocus`. While `autofocus` can easily be misused, we want to rely on other tools like an axe rule to surface those issues rather than code linting. This PR makes it so we turn the [jsx-a11y/no-autofocus](https://github.com/jsx-eslint/eslint-plugin-jsx-a11y/blob/main/docs/rules/no-autofocus.md) rule which is on as part of the recommended rulesets, off by default.


dgreif (approved):

dgreif:
I've published `v4.8.0` with this change. @khiga8 do you mind bumping the dependency in `github/github`?

PR #446, opened 2023-07-05
create rule: github/a11y-no-visually-hidden-interactive-element

kendallgassner:
### What

relates to: https://github.com/github/accessibility/issues/3739

This was an interesting lint rule to write because we not only utilize the `className='sr-only'` to visually hide react content but we use the Primer component `VisuallyHidden`.

I wrote an eslint rule for both scenarios. This eslint rule determines if a `JSX.Elements` is visually hidden (i.e. has the `sr-only` className or of type `VisuallyHidden`). If it the element is visually hidden then it checks its self and children to see if it contains and interactive elements. 

#### What is an interactive element?
I hard coded a list of interactive elements to check against: 
- `a`
- `button`
- `summary`
- `select`
- `option`
- `textarea`

I thought it was better to have a `dis-allow` list rather then an allow list (e.g. `span`, `div` etc.)

#### What is considered invalid?
- nested interative elements:`<VisuallyHidden><div><button>Submit</button></div></VisuallyHidden>`
- as props: `<VisuallyHidden as='button'>Submit</VisuallyHidden>`
- multiple classNames: `<button className='other sr-only'>Submit</button>`


lib/index.js
```diff
@@ -1,5 +1,6 @@
 module.exports = {
   rules: {
+    'a11y-no-sr-only-class-when-focusable': require('./rules/a11y-no-sr-only-class-when-focusable'),
```
kendallgassner, line 3: Should we wait to include this until: https://github.com/github/accessibility/issues/3868

khiga8: I think it's fine! We could always override disable is temporarily in our projects.

khiga8 (commented):
Yay I'm excited to see this! I will review more in-depth tomorrow but leaving a comment per quick skim of the description!

> but we use the Primer component VisuallyHidden

TIL about `VisuallyHidden` in Primer! 🤯

Since `sr-only` and `VisuallyHidden` is very GitHub/Primer-centric, and many consumers use this plugin outside of GitHub/Primer-consumers, could we allow this (the sr-only classname and the type) to be configured through the ESLint rule-level configs? I believe there are examples of this in the `jsx-a11y` library.

For example, if I am a consumer whose project uses the class `visually-hidden` instead of `sr-only`, or the type `ScreenReaderOnly` instead of `VisuallyHidden`, I'd still like to be able to benefit from this rule.

kendallgassner:
@khiga8 Love that thought! Great idea... I'll add this in and re-tag you for an in-depth review when its updated.

kendallgassner:
Interestingly, my passion week project is a use-case where we do want to visuallyHide an interactive element 🤔

khiga8:
@kendallgassner Could you bring up this example in the channel so we can discuss if it's valid?

lib/rules/a11y-no-visually-hidden-interactive-element.js
```diff
@@ -0,0 +1,90 @@
+const {getProp, getPropValue} = require('jsx-ast-utils')
+const {getElementType} = require('../utils/get-element-type')
+const {generateObjSchema} = require('eslint-plugin-jsx-a11y/lib/util/schemas')
+
+const defaultClassName = 'sr-only'
```
kendallgassner, line 5: considering making the default `visually-hidden` for readability.

tests/a11y-no-visually-hidden-interactive-element.js
```diff
@@ -0,0 +1,45 @@
...
+ruleTester.run('a11y-no-visually-hidden-interactive-element', rule, {
+  valid: [
+    {code: '<VisuallyHidden as="h2">Submit</VisuallyHidden>'},
+    {code: "<div className='sr-only'>Text</div>;"},
+    {code: '<VisuallyHidden><div>Text</div></VisuallyHidden>'},
+    {code: "<div className='other visually-hidden'>Text</div>;"},
+    {code: "<span className='sr-only'>Text</span>;"},
+    {code: "<button className='other'>Submit</button>"},
+    {code: "<input className='sr-only' />"},
+    {code: "<a className='other show-on-focus'>skip to main content</a>"},
```
kendallgassner, line 26: Check that `show-on-focus` class is valid! This way I don't have to remove the `a` tag from the list.

lib/rules/a11y-no-visually-hidden-interactive-element.js
khiga8: I prefer `sr-only` just because it's more widely used within different GitHub projects! As long as we document what the default is and how consumers can change it, I think `sr-only` would be okay.

docs/rules/a11y-no-visually-hidden-interactive-element.md
```diff
@@ -0,0 +1,55 @@
+# Ensures that interactive elements are not visually hidden (`github/a11y-no-visually-hidden-interactive-element`)
+
+<!-- end auto-generated rule header -->
+
+## Rule Details
+
+This rule guards against visually hiding interactive elements. If a sighted keyboard user navigates to an interactive element that is visually hidden they might become confused and assume that keyboard focus has been lost.
```
khiga8, line 7: Can we note the default class/type, as well as which interactive elements we're checking against?

I like the note you added in the code about why we're excluding `input`. Could we add it here too?

lib/rules/a11y-no-visually-hidden-interactive-element.js
```diff
@@ -0,0 +1,90 @@
...
+
+/** Note: we are not including input elements at this time
+ * because a visually hidden input field might cause a false positive.
+ * (e.g. fileUpload https://github.com/primer/react/pull/3492)
+ */
+const INTERACTIVELEMENTS = ['a', 'button', 'summary', 'select', 'option', 'textarea']
+
+const checkIfInteractiveElement = (context, node) => {
+  const elementType = getElementType(context, node.openingElement)
+  const asProp = getPropValue(getProp(node.openingElement.attributes, 'as'))
```
khiga8, line 21: I love this approach of just directly grabbing the tag from `as`, a lot more than this experimental approach I added some time ago! I've been wanting to revisit and remove this - [Component mapping experimental](https://github.com/github/eslint-plugin-github#component-mapping-experimental)! 

Since `as` is Primer React specific, can we allow the `as` to be configured in the config as well? (e.g. maybe some design systems use `tag` or `component`). 

I think this line can be moved into `getElementType`, and we can get rid of the comopnent mappign stuff (lines 15-34)!

khiga8 (commented):
Let's add this rule to [react.js](https://github.com/github/eslint-plugin-github/blob/main/lib/configs/react.js) so it's on by default :) 

I have a few suggestions. This is coming along really nicely <3

lib/rules/a11y-no-visually-hidden-interactive-element.js
khiga8: Maybe we can pair on this in our sync :)

kendallgassner: `htmlPropName` - name for 'as' config

theinterned (approved):
I will definitely defer to @khiga8 's review here, but no concerns form the web-systems standpoint

kendallgassner:
@khiga8 I added the documentation updates and the `htmlPropName`. Can I clean up [Component mapping experimental](https://github.com/github/eslint-plugin-github#component-mapping-experimental)! and add this to `getElementType` in a follow up pr? -- I think it would be nice to test those changes and look at the logic in a clean state/

tests/a11y-no-visually-hidden-interactive-element.js
```diff
@@ -0,0 +1,45 @@
...
+    ecmaFeatures: {
+      jsx: true,
+    },
+  },
+})
+
+const errorMessage =
+  'Avoid visually hidding interactive elements. Visually hiding interactive elements can be confusing to sighted keyboard users as it appears their focus has been lost when they navigate to the hidden element.'
+
+ruleTester.run('a11y-no-visually-hidden-interactive-element', rule, {
```
khiga8, line 17: Can we add a test to make sure the options work as expected? [Here's an example](https://github.com/jsx-eslint/eslint-plugin-jsx-a11y/blob/f0d2ddb65f21278ad29be43fb167a1092287b4b1/__tests__/src/rules/anchor-ambiguous-text-test.js#L112)! :)

khiga8 (approved):
I have one test suggestion, but please feel free to merge after that!

This looks great 🚀! Good call on doing the other updates in a separate PR, I agree!

tests/a11y-no-visually-hidden-interactive-element.js
kendallgassner: Amazing! I didn't realize that was a test option!

PR #448, opened 2023-07-10
Bump aria-query and update to fix tests

khiga8:
Follow-up to https://github.com/github/eslint-plugin-github/pull/447#issuecomment-1629097355
Fixes: https://github.com/github/accessibility/issues/3966

[Related slack thread](https://github.slack.com/archives/C0FSWLQ0Y/p1688483187990509)

We see breaking tests for our `role-supports-aria-props` rule with the latest `aria-query` bump. This is because the latest `aria-query` bump includes update to the element to role mapping.

It took me a bit to fully understand the logic in our `role-supports-aria-props` rule. In short, we're:
1. Creating a new object map based off the map returned by [elementRoleMap-test](https://github.com/A11yance/aria-query/blob/main/__tests__/src/elementRoleMap-test.js). We remove some constraints from the key which isn't useful to us.
2. We take the node we are given, and construct a new `key`.
3. With this `key`, we check if there's a match in the object map we constructed.
4. If there isn't a match, the role is undefined, and we just skip the node in the rule. 
5. If there is a match, we check if there are any prohibited attributes being used.

It looks like our current code does not account for the various new constraints introduced in the latest `aria-query`. This results in incorrect roles being returned, and the tests for our rule breaking. In addition, some of the key to role mappings have changed.

I recommend reviewing by commit. 

The changes in this PR include:

* Don't delete the attributes `constraints` because it will influence the mapping output.
* Remove `aria-expanded` from the disambiguating attributes on `summary`. This has been removed in [these lines of code](https://github.com/A11yance/aria-query/pull/447/files#diff-250492ccb7c3a2082cc9fc09e5ac12a5775cd0733438c2acbe32276612258bb0L40-L50).
* Update the disambiguating attributes to include `aria-label` and `aria-labelledby` (restricted to `section`, `aside`)
* Update the disambiguating attributes to include `name` restricted to `form`
* Update the disambiguating attributes to include `alt` restricted to `img`
* Update tests based on changes in `aria-query`.
* Extract logic into a `getRole` helper and add tests.

**In a follow-up PR, I will rename this rule to `a11y-role-supports-aria-props` to ensure it's tracked by the scorecard.**

#### Other thoughts

Interestingly, the `jsx-a11y` library hard-codedly defines the logic for a lot of tags in [implicitRoles](https://github.com/jsx-eslint/eslint-plugin-jsx-a11y/tree/main/src/util/implicitRoles). The downside of that approach is that it gets out of date, and isn't utilize the mapping in `aria-query` which gets more frequently updated.

The downside of the approach we're taking by constructing a key/map and using the mapping as the source of truth, is that the key we're constructing might stop returning a match depending on updates to `aria-query`. For example, more recent aria-query updates introduced a bunch of constraints which previously weren't there. I tried to protect against brittleness by adding more tests for `getRole` which I've extracted. I think we can revisit this later.


lib/rules/role-supports-aria-props.js
```diff
@@ -9,7 +9,10 @@ const elementRolesMap = new ObjectMap()
 for (const [key, value] of elementRoles.entries()) {
   // - Remove unused `constraints` key
   delete key.constraints
-  key.attributes = key.attributes?.filter(attribute => !('constraints' in attribute))
+  // - Remove `constraints` key within `attributes`
+  for (const attribute of key.attributes || []) {
+    delete attribute.constraints
+  }
```
khiga8, line 15: We probably don't want to be deleting constraints at all here.

tests/role-supports-aria-props.js
```diff
@@ -57,9 +57,6 @@ ruleTester.run('role-supports-aria-props', rule, {
     {code: '<a href="#" aria-owns />'},
     {code: '<a href="#" aria-relevant />'},
 
-    // this will have global
-    {code: '<a aria-checked />'},
```
khiga8, line 61: This is now treated as `generic`

tests/role-supports-aria-props.js
```diff
@@ -78,30 +75,6 @@ ruleTester.run('role-supports-aria-props', rule, {
     {code: '<area href="#" aria-owns />'},
     {code: '<area href="#" aria-relevant />'},
 
-    // this will have global
-    {code: '<area aria-checked />'},
-
-    // LINK TESTS - implicit role is `link`
```
khiga8, line 84: `<link>` maps to nothing!

tests/role-supports-aria-props.js
```diff
@@ -414,6 +392,10 @@ ruleTester.run('role-supports-aria-props', rule, {
     },
     {
       code: '<section aria-expanded />',
+      errors: [getErrorMessage('aria-expanded', 'generic')],
+    },
+    {
+      code: '<section aria-label="something" aria-expanded />',
```
khiga8, line 398: If `<section aria-label>`, we treat it as `region`.

khiga8:
Please give an initial review if you have capacity! Since this was quite hard for me to dive into, I will review again with fresh eyes tomorrow.

tests/utils/get-element-type.js
```diff
@@ -1,34 +1,11 @@
 const {getElementType} = require('../../lib/utils/get-element-type')
+const {mockJSXAttribute, mockJSXOpeningElement} = require('./helpers')
+
 const mocha = require('mocha')
 const describe = mocha.describe
 const it = mocha.it
 const expect = require('chai').expect
 
-function mockJSXAttribute(prop, propValue) {
```
khiga8, line 7: I'm extracting this into a helper.

khiga8:
Interestingly, jsx-a11y library hard-codedly defines the logic for a lot of tags in [implicitRoles](https://github.com/jsx-eslint/eslint-plugin-jsx-a11y/tree/main/src/util/implicitRoles).

lib/utils/get-role.js
```diff
@@ -0,0 +1,93 @@
+const {getProp, getPropValue} = require('jsx-ast-utils')
```
khiga8, line 1: I would appreciate @smockle eyes on this!

lib/utils/get-role.js
```diff
@@ -0,0 +1,93 @@
...
+const {elementRoles} = require('aria-query')
+const {getElementType} = require('./get-element-type')
+const ObjectMap = require('./object-map')
+
+// Clean-up `elementRoles` from `aria-query`
+const elementRolesMap = new ObjectMap()
```
khiga8, line 7: This code is constructing a cleaned up version of [elementRoles](https://github.com/A11yance/aria-query/blob/main/__tests__/src/elementRoleMap-test.js#L4). Below, we'll be constructing a key and grabbing the value from `elementRoles`.

khiga8:
There won't always be a matching `role` returned but I think that's fine, as long as we cover the majority of cases.

jfuchs (approved):
Thank you for this! If you haven't already, it might be nice to verify in a branch that if you rebase [this dependabot update](https://github.com/github/eslint-plugin-github/pull/452) on top of this this PR, tests pass.

khiga8:
@jfuchs this PR already includes the same `aria-query` bump as in the dependabot PR, so I think we should be good!!

lib/utils/get-role.js
```diff
@@ -0,0 +1,93 @@
...
+for (const [key, value] of elementRoles.entries()) {
+  // - Remove unused `constraints` key
+  delete key.constraints
```
smockle, line 10: Is this line still necessary?

khiga8: Oooh nice catch! It looks like removing this constraints can result in incorrect mappings in some scenarios... 

For example, `elementRoles` includes:
```
[{"constraints": ["ancestor table element has table role"], "name": "td"}, ["cell"]]
[{"constraints": ["ancestor table element has grid role", "ancestor table element has treegrid role"], "name": "td"}, ["gridcell"]]
```

If we remove the constraints key, we're left with:

```
[{"name": "td"}, ["cell"]]
[{"name": "td"}, ["gridcell"]]
```

Since we cannot always accurately parse out the role of an ancestor in a linter context, I think we'd want to return undefined instead of potentially returning a false positive role.

I will follow-up!

khiga8: I deleted this line in https://github.com/github/eslint-plugin-github/pull/448/commits/8fb64196e8aa8df8819facda7284100f2577a30a!

However, there were two scenarios where it seemed appropriate to assume the role despite the constraints defined in `aria-query`. I opted to hard-code those with a comment on why! You can see the commit for more information.

I've added additional tests in https://github.com/github/eslint-plugin-github/pull/448/commits/aaca700b73ca9bc778bf254defc78eecd7e6e1e9#diff-a6b7fa0d4a2be411ccfcc6e6db5466c135100a1a045599f5c211bb91c37b839dR180-R190.

khiga8:
I moved the map constructing code into a function with additional code comments in https://github.com/github/eslint-plugin-github/pull/448/commits/40c0b2b7a518b36b238adebae6911f0cee1156d4.

khiga8:
I will merge this PR and https://github.com/github/eslint-plugin-github/pull/454 in tomorrow morning (EST)!!!

PR #450, opened 2023-07-11
Adds `svg-has-accessible-name` rule

lindseywild:
Relates to https://github.com/github/accessibility/issues/3972.

### Description

Adds a lint rule to ensure SVGs have an accessible name by checking for a title element as the first descendant of the `<svg>` element or an `aria-label` or `aria-labelledby`. It also checks if the element is presentational, in which an error won't be thrown.


lib/rules/a11y-svg-has-accessible-name.js
```diff
@@ -0,0 +1,41 @@
...
+  create(context) {
+    return {
+      JSXOpeningElement: node => {
+        const elementType = getElementType(context, node)
+
+        // Check if there is a nested title element that is the first child of the `<svg>`
+        const hasNestedTitleAsFirstChild =
+          node.parent.children.length &&
+          node.parent.children[0].type === 'JSXElement' &&
+          node.parent.children[0].openingElement.name.name === 'title'
```
keithamus, line 22: May want to guard these which will prevent errors on property access:

```suggestion
        const hasNestedTitleAsFirstChild =
          node.parent.children?.[0]?.type === 'JSXElement' &&
          node.parent.children?.[0]?.openingElement?.name?.name === 'title'
```

lib/rules/a11y-svg-has-accessible-name.js
```diff
@@ -0,0 +1,41 @@
...
+      description: 'SVGs must have an accessible name',
+      url: require('../url')(module),
+    },
+    schema: [],
+  },
+
+  create(context) {
+    return {
+      JSXOpeningElement: node => {
+        const elementType = getElementType(context, node)
```
keithamus, line 16: An early return here will speed up execution which can get bogged down with many small cuts:

```suggestion
        const elementType = getElementType(context, node)
        if (elementType !== 'svg') return
```

keithamus (approved):
Very nice :+1:

tests/a11y-svg-has-accessible-name.js
```diff
@@ -0,0 +1,49 @@
+const rule = require('../lib/rules/a11y-svg-has-accessible-name')
```
khiga8, line 1: nice tests!! thank you!!

docs/rules/a11y-svg-has-accessible-name.md
```diff
@@ -0,0 +1,73 @@
...
+
+💼 This rule is enabled in the ⚛️ `react` config.
+
+<!-- end auto-generated rule header -->
+
+## Rule Details
+
+An `<svg>` must have an accessible name. Set `aria-label` or `aria-labelledby`, or nest a `<title>` element as the first child of the `<svg>` element.
+
+However, if the `<svg>` is purely decorative, hide it with `aria-hidden="true"` or `role="presentation"`.
```
khiga8, line 11: Ahhh I didn't consider `role="presentation"`! TIL!

khiga8 (approved):
This looks great! Let's ship ittttt 🚀

PR #453, opened 2023-07-11
Create rule: a11y-no-title-attribute

kendallgassner:
### What

A new rule to guard against the `title` attribute. 

The title attribute is an inaccessible version of a tooltip. It is often used to provide context when text is truncated or add additional context to an interactive element like a link. This practice is exclusionary and prevents keyboard and mobile users from accessing information.


docs/rules/a11y-no-title-attribute.md
```diff
@@ -0,0 +1,43 @@
+# Guards against developers using the title attribute (`github/a11y-no-title-attribute`)
+
+<!-- end auto-generated rule header -->
+
+The title attribute is strongly discouraged. The only exception is on an `<iframe>` element. It is hardly useful and cannot be accessed by multiple groups of users including keyboard-only users and mobile users.
```
kendallgassner, line 5: Most of this is directly from the [rails rule](https://github.com/github/erblint-github/blob/main/docs/rules/accessibility/no-title-attribute.md). With the exception of the last sentence.

lib/rules/a11y-no-visually-hidden-interactive-element.js
```diff
@@ -76,7 +76,6 @@ module.exports = {
             message:
               'Avoid visually hidding interactive elements. Visually hiding interactive elements can be confusing to sighted keyboard users as it appears their focus has been lost when they navigate to the hidden element.',
           })
-          return
```
kendallgassner, line 79: just unnessary.

khiga8:
I wonder if `title` is ever used in Primer React as a prop that means something else 🤔, so we don't accidentally falsely flag it. 

Do you know @tylerjdev?

khiga8 (approved):
So speeedy! This looks good to me! I have one question, which maybe @tylerjdev knows the answer to?

TylerJDev:
@khiga8 `SelectPanel` uses `title` for the heading, as well as `[ActionList` for one of the subcomponents](https://primer.style/react/ActionList#actionlistgroup). Not sure if there are more instances, but I can gather any additional components that use `title` as a prop if it's helpful!

khiga8:
Thanks @tylerjdev! I wonder what the best approach would be to minimize false positives?

We could start conservatively and only run this check on HTML elements, if there's a way to identify that easily (all lower case?). 

Alternatively, we could create an ignore-list that can be configured.

What do you think @tylerjdev and @kendallgassner?

kendallgassner:
@khiga I lean towards a list of semantic html elements -- its less work to manage.

TylerJDev:
I think I prefer the first option of only running this on native HTML elements. I thinking for an ignore list, that should primarily be Primer's responsibility, and might be hard to maintain if we hade on in this repo. 

I do wonder if it'd make sense to make a duplicate rule in `eslint-plugin-primer-react` that would handle the same logic as this rule, but only for Primer components? There's a decent bit of [`title` usage on components that do not support this](https://musical-adventure-wlr3n3k.pages.github.io/?attribute=%22title%22) as a custom prop, so it could be worth it, but would love to know any additional thoughts.

khiga8:
@tylerjdev That sounds good. I think starting with HTML only will minimize false positives so I like that too. 

> I do wonder if it'd make sense to make a duplicate rule in eslint-plugin-primer-react that would handle the same logic as this rule, but only for Primer components?

If we do introduce an ignore list config for this rule, it could be maintained by `eslint-plugin-primer-react` in [recommended.js](https://github.com/primer/eslint-plugin-primer-react/blob/main/src/configs/recommended.js#L17), so we wouldn't need to duplicate the rule. `eslint-plugin-primer-react` already maintains configs for this library (though it's out of date right now).

Note: the config in `eslint-plugin-primer-react` is out of date and should to be updated as part of https://github.com/primer/eslint-plugin-primer-react/issues/56.

PR #459, opened 2023-07-17
Re-introduce accidentally removed testing commands

khiga8:
A few commands were unnintentionaly removed in [this PR](https://github.com/github/eslint-plugin-github/pull/449/files#diff-7ae45ad102eab3b6d7e7896acd08c427a9b25b346470d7bc6507b6481575d519L18), so adding it back!

There is a breaking docs lint rule which wasn't caught before, so fixed it in this PR.


package.json
```diff
@@ -15,7 +15,7 @@
     "lint:eslint-docs": "npm run update:eslint-docs -- --check",
     "lint:js": "eslint .",
     "pretest": "mkdir -p node_modules/ && ln -fs $(pwd) node_modules/",
-    "test": "mocha tests/**/*.js tests/",
+    "test": "npm run eslint-check && npm run lint && mocha tests/**/*.js tests/",
```
kendallgassner, line 18: Apologies!

kendallgassner (approved):

PR #460, opened 2023-07-17
Lock `aria-query` and set config override for false positive rule

khiga8:
In the latest release of eslint-plugin-github which includes a bump of [aria-query 5.3.0](https://github.com/github/eslint-plugin-github/pull/448), one of the `jsx-a11y` rule is newly flagging:

```tsx
<td role="cell">
```

This seems to be unexpected.

We depend on `aria-query` 5.3.0 in this project for one of our rules, but this is higher than what the latest release of eslint-plugin-jsx-a11y depends on. I think this is resulting in us seeing unexpected behavior which would normally be taken care of by the `eslint-plugin-jsx-a11y` library. This PR:

* temporarily overrides the config for the rule to avoid raising a false positive.
* locks the aria-query version so we avoid running into situations in the future where `eslint-plugin-jsx-a11y` hasn't caught up with changes in aria-query.

Although I considered reverting 5.3.0, I opted not to since we made a ton of updates in our code in https://github.com/github/eslint-plugin-github/pull/448 in the process of bumping to 5.3.0, and I don't think it's worth reverting.


kendallgassner (approved):
Makes sense thanks for leaving the todo

PR #461, opened 2023-07-17
Fix bugs with `getRole` and `getElementType`

khiga8:
We noticed a few unexpected behavior with the latest eslint-plugin-github 4.9.0 release, which all seems to stem from using `getPropValue` instead of `getLiteralPropValue` in our methods `getElementType` and `getRole`.

`getElementType` will return a div for the following. It should instead return `Box` and not try to assess the conditional.

```tsx
<Box as={isNavigationOpen ? 'div' : 'nav'} />
```

-- 

`getRole` will return a "generic" here. I don't think it should return anything.

```tsx
<Box role={isNavigationOpen ? 'generic' : 'navigation'} />
```
-- 

`getRole` will return a "role" here. It should retur undefined since this is `role` is set to a variable.

```tsx
<Box role={role} />
```
-- 

This PR:

* adds test which initially fail to reflect the failures we observed
* adds fix for test which is to swap `getPropValue` with `getLiteralPropValue`.

### DIFF
--- a/lib/utils/get-element-type.js
+++ b/lib/utils/get-element-type.js
@@ -1,4 +1,4 @@
-const {elementType, getProp, getPropValue} = require('jsx-ast-utils')
+const {elementType, getProp, getLiteralPropValue} = require('jsx-ast-utils')
 
 /*
 Allows custom component to be mapped to an element type.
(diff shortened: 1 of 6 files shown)


lib/utils/get-element-type.js
```diff
@@ -12,7 +12,7 @@ function getElementType(context, node, ignoreMap = false) {
 
   // check if the node contains a polymorphic prop
   const polymorphicPropName = settings?.github?.polymorphicPropName ?? 'as'
-  const rawElement = getPropValue(getProp(node.attributes, polymorphicPropName)) ?? elementType(node)
+  const rawElement = getLiteralPropValue(getProp(node.attributes, polymorphicPropName)) ?? elementType(node)
```
kendallgassner, line 15: 👑

kendallgassner (approved):
Good catch ❤️

PR #466, opened 2023-07-18
Use `getLiteralPropValue` for sr-only class

khiga8:
We caught a scenario where `getPropValue(getProp(node.openingElement.attributes, 'className'))` returns a function.

Example code:

```tsx
<Foo className={({isOn}) => { return isOn || isOnProps ? `${className} selected`.trim() : className ?? ''}}/>
```

Since `getPropValue` will return a function in this case, the `classes.includes` call ends up throwing an error.

This PR updates the logic to use `getLiteralPropValue`and moves around the logic.


TylerJDev (approved):
Good catch! 🎉

srt32 (approved):

PR #503, opened 2024-02-03
Update no-then.md

cefn:
The 'improved code' after linting was unnecessarily complex introducing a named variable across multiple lines for no reason which made it look better to keep using `then`. A bit surprising.

Previous catch example was code which didn't differentiate between failing case (which is now a deliberate null) and undefined case (which is often the result of an error).

Also the previous catch example was the only example for a rule which is called 'no-then' and therefore the failing example didn't use 'then' either.

### DIFF
--- a/docs/rules/no-then.md
+++ b/docs/rules/no-then.md
@@ -35,7 +35,7 @@ function getProcessedData(url) {
 async function countProcessedData(url) {
   const data = await downloadData(url);
   return data.length
-```
+}
 
 ```js
 async function getProcessedData(url) {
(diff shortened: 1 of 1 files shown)


theinterned (approved):
Thank you. I agree these examples are an improvement!

docs/rules/no-then.md
```diff
@@ -12,26 +12,39 @@ https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Statements/asy
...
 }
 ```
 
 👍 Examples of **correct** code for this rule:
 
+```js
+async function countProcessedData(url) {
+  const data = await downloadData(url);
+  return data.length
+}
```
theinterned, line 38: Oh no! now we lost the closing fence 🤦 

```suggestion
```js
async function countProcessedData(url) {
  const data = await downloadData(url);
  return data.length
}
```
```

PR #525, opened 2024-05-16
Fix bug in `getElementType` logic 

khiga8:
This PR fixes a few issues:

#### 1. Fixing bug in `getElementType` logic

```.ts
const mapping = {
  'Box': 'div'
}
```

```.tsx
<Box as={something} />
```

will be interpreted as `div`. Currently the linter will check the mapping if it is unable to interpret the polymorphic value. However, it should not do that because that is not a correct interpretation.

#### 2. Fixing mjs tests that aren't running

I noticed the tests in `utils` aren't running even when calling `npm run test`.

This seems related to [this change](https://github.com/github/eslint-plugin-github/commit/5c76fdec19b2f9cbed3856f6fbdf59fdc85ad7ff) where the tests were converted to `.mjs`, but the `.mjs` tests aren't actually running. When I fixed the command, the tests are being targeted. The test will not run properly without the `.js` extension.  

#### 3. Drops node 14 support

Drops node 14 support per [slack thread](https://github.slack.com/archives/CMZ4DC9BL/p1715874890704489)!

#### 4. Adds node 16 support

### DIFF
--- a/.eslintrc.js
+++ b/.eslintrc.js
@@ -10,6 +10,7 @@ module.exports = {
   extends: [require.resolve('./lib/configs/recommended'), 'plugin:eslint-plugin/all'],
   plugins: ['eslint-plugin'],
   rules: {
+    'import/extensions': 'off',
     'import/no-commonjs': 'off',
     'filenames/match-regex': 'off',
     'i18n-text/no-en': 'off',
--- a/.github/workflows/nodejs.yml
+++ b/.github/workflows/nodejs.yml
@@ -12,7 +12,7 @@ jobs:
 
     strategy:
       matrix:
-        node-version: [16, 18]
+        node-version: [18]
 
     steps:
       - uses: actions/checkout@v4
--- a/.github/workflows/nodejs.yml
+++ b/.github/workflows/nodejs.yml
@@ -12,7 +12,7 @@ jobs:
 
     strategy:
       matrix:
-        node-version: [18]
+        node-version: [18, 20]
 
     steps:
       - uses: actions/checkout@v4
--- a/.github/workflows/nodejs.yml
+++ b/.github/workflows/nodejs.yml
@@ -12,7 +12,7 @@ jobs:
 
     strategy:
       matrix:
-        node-version: [14, 16, 18]
+        node-version: [16, 18]
 
     steps:
       - uses: actions/checkout@v4
--- a/lib/utils/get-element-type.js
+++ b/lib/utils/get-element-type.js
@@ -17,15 +17,20 @@ function getElementType(context, node, lazyElementCheck = false) {
   // check if the node contains a polymorphic prop
   const polymorphicPropName = settings?.github?.polymorphicPropName ?? 'as'
 
-  const rawElement = elementType(node)
+  let checkConditionalMap = true
 
-  if (getProp(node.attributes, polymorphicPropName)) {
-    return getLiteralPropValue(getProp(node.attributes, polymorphicPropName)) ?? rawElement
-  } else if (settings?.github?.components?.[rawElement]) {
-    return settings.github.components[rawElement]
-  } else {
-    return rawElement
+  const prop = getProp(node.attributes, polymorphicPropName)
+  const literalPropValue = getLiteralPropValue(getProp(node.attributes, polymorphicPropName))
+  if (prop && !literalPropValue) {
+    checkConditionalMap = false
   }
+  const rawElement = getLiteralPropValue(getProp(node.attributes, polymorphicPropName)) ?? elementType(node)
+
+  // if a component configuration does not exists, return the raw element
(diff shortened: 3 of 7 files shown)


lib/utils/get-element-type.js
```diff
@@ -16,15 +16,16 @@ function getElementType(context, node, lazyElementCheck = false) {
...
-  const rawElement = getLiteralPropValue(getProp(node.attributes, polymorphicPropName)) ?? elementType(node)
 
-  // if a component configuration does not exists, return the raw element
-  if (!settings?.github?.components?.[rawElement]) return rawElement
+  const rawElement = elementType(node)
 
-  const defaultComponent = settings.github.components[rawElement]
-
-  // check if the default component is also defined in the configuration
-  return defaultComponent ? defaultComponent : defaultComponent
```
khiga8, line 27: Was this intentional @kendallgassner?

package.json
```diff
@@ -15,7 +15,7 @@
     "lint:eslint-docs": "npm run update:eslint-docs -- --check",
     "lint:js": "eslint .",
     "pretest": "mkdir -p node_modules/ && ln -fs $(pwd) node_modules/",
-    "test": "npm run eslint-check && npm run lint && mocha tests/**/*.js tests/",
+    "test": "npm run eslint-check && npm run lint && mocha tests/**/*.js tests/**/*.mjs",
```
khiga8, line 18: The `mjs` tests aren't actually running since [this update](https://github.com/github/eslint-plugin-github/commit/5c76fdec19b2f9cbed3856f6fbdf59fdc85ad7ff).

khiga8:
@github/web-systems-reviewers 

Node 14 builds seem to be failing. Can we drop support?

.eslintrc.js
```diff
@@ -10,6 +10,7 @@ module.exports = {
   extends: [require.resolve('./lib/configs/recommended'), 'plugin:eslint-plugin/all'],
   plugins: ['eslint-plugin'],
   rules: {
+    'import/extensions': 'off',
```
khiga8, line 13: In order for the JS files to be loaded into `.mjs`, we need an extension.

This library seems to be using a mix of module JS and common JS now, so we'll want to turn this rule off.

.github/workflows/nodejs.yml
```diff
@@ -12,7 +12,7 @@ jobs:
 
     strategy:
       matrix:
-        node-version: [14, 16, 18]
+        node-version: [18]
```
JoyceZhu, line 15: Should we add 20? I've been seeing more Actions migrate to it, and it's now a year old; if nothing breaks might be nice to start testing against it.

(disclaimer: I'm not the expert on node versions.)

khiga8: Ooh okay! let's see what it tells us!

tests/utils/get-element-type.mjs
```diff
@@ -63,4 +60,15 @@ describe('getElementType', function () {
     ])
     expect(getElementType({}, node)).to.equal('Box')
   })
+
+  it('returns raw type when polymorphic prop is set to non-literal expression even with component setting', function () {
+    // <Box as={isNavigationOpen ? 'generic' : 'navigation'} />
```
JoyceZhu, line 65: remove the comment out?

khiga8, line 65:
```suggestion
```

.github/workflows/nodejs.yml
khiga8: WOO! Added, and passed!

package.json
JoyceZhu: > This library seems to be using a mix of module JS and common JS now

I wonder if that's why? 🫤

tests/utils/get-element-type.mjs
```diff
@@ -1,10 +1,7 @@
 import {expect} from 'chai'
-import {getElementType} from '../../lib/utils/get-element-type'
-import {mockJSXAttribute, mockJSXConditionalAttribute, mockJSXOpeningElement} from './mocks'
-
-const mocha = require('mocha')
-const describe = mocha.describe
-const it = mocha.it
+import {getElementType} from '../../lib/utils/get-element-type.js'
+import {mockJSXAttribute, mockJSXConditionalAttribute, mockJSXOpeningElement} from './mocks.js'
```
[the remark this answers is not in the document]
JoyceZhu: > This library seems to be using a mix of module JS and common JS now

I wonder if that's why? 🫤

JoyceZhu (approved):

lib/utils/get-element-type.js
kendallgassner: definitely not

dgreif (approved):
👍 for web-systems-reviewers

TylerJDev (approved):

PR #531, opened 2024-06-04
Provide `no-redundant-roles` exception for `rowgroup`

khiga8:
Relates to: https://github.com/github/accessibility/issues/5304

Follow-up to: https://github.com/github/accessibility/discussions/4921

## What
This PR adds an exception so that the `no-redundant-roles` rule does not flag:
*  `<thead role="rowgroup">`
*  `<tbody role="rowgroup">`

## Why
We have determined this markup is necessary in some scenarios to improve screen reader semantics, and should not be flagged as redundant. Because these are the only reported false positive instances, it does not seem worth turning the rule off completely. If there are other special scenarios reported, we should add it as an exception here in this config.

See [specific example in primer/react](https://github.com/primer/react/blob/8cd6007e589d33fd89a1ab2a93547b9977e274b2/src/DataTable/Table.tsx#L298).

### DIFF
--- a/lib/configs/react.js
+++ b/lib/configs/react.js
@@ -42,5 +42,13 @@ module.exports = {
         fieldset: ['radiogroup', 'presentation'],
       },
     ],
+    'jsx-a11y/no-redundant-roles': [
+      'error',
+      {
+        nav: ['navigation'], // default in eslint-plugin-jsx-a11y
+        tbody: ['rowgroup'],
+        thead: ['rowgroup'],
+      },
+    ],
   },
 }


lib/configs/react.js
```diff
@@ -42,5 +42,13 @@ module.exports = {
         fieldset: ['radiogroup', 'presentation'],
       },
     ],
+    'jsx-a11y/no-redundant-roles': [
+      'error',
+      {
+        nav: ['navigation'], // default in eslint-plugin-jsx-a11y
```
TylerJDev, line 48: Just curious, is this needed because the object is overridden when we provide our own config/mapping? 🤔

TylerJDev (approved):
Looks good, thanks a ton too! ✨

lib/configs/react.js
khiga8: Yes, that is correct!

khiga8: Or at least thats how I remember it.. let me double check 👀

khiga8: If I'm interpreting the [logic](https://github.com/jsx-eslint/eslint-plugin-jsx-a11y/blob/main/src/rules/no-redundant-roles.js#L60) correctly, it looks like the rule will use the default `{ nav: ['navigation']}` _if_ the consumer isn't already passing an alternate option! This seems fine to keep!

PR #563, opened 2024-10-30
update node variant

arelia:


.devcontainer/Dockerfile
```diff
@@ -1,7 +1,7 @@
 # See here for image contents: https://github.com/microsoft/vscode-dev-containers/tree/v0.222.0/containers/javascript-node/.devcontainer/base.Dockerfile
 
 # [Choice] Node.js version (use -bullseye variants on local arm64/Apple Silicon): 16, 14, 12, 16-bullseye, 14-bullseye, 12-bullseye, 16-buster, 14-buster, 12-buster
-ARG VARIANT="16"
+ARG VARIANT="22"
 FROM mcr.microsoft.com/vscode/devcontainers/javascript-node:0-${VARIANT}
```
gracepark, line 5:
```suggestion
# [Choice] Node.js version (use -bullseye variants on local arm64/Apple Silicon): 20, 18, 20-bullseye, 18-bullseye
ARG VARIANT="20"
FROM mcr.microsoft.com/vscode/devcontainers/javascript-node:0-${VARIANT}
```

gracepark (commented):
I think we need to go to version 20:

https://mcr.microsoft.com/v2/devcontainers/javascript-node/tags/list

That means we'll need to update the devcontainer.json

.devcontainer/Dockerfile
gracepark: Here's the other change: https://github.com/github/eslint-plugin-github/pull/564/files

PR #564, opened 2024-10-30
Update devcontainer.json

gracepark:
Basing it off of: https://github.com/github/eslint-plugin-github/pull/563

This seems to work for my codespace


arelia (approved):
looks good as this allows codespace to build

PR #565, opened 2024-10-30
Update Dockerfile

gracepark:


arelia (approved):
looks good as using `22` vs `0-22` allows codespace to build

`0-22` is [not available at this time](https://mcr.microsoft.com/v2/devcontainers/javascript-node/tags/list) and causes the error: `Error: Command failed: docker inspect --type image mcr.microsoft.com/vscode/devcontainers/javascript-node:0-22`

PR #567, opened 2024-10-31
fix: async-currenttarget/preventdefault doesn’t consider nested scopes

kurtextrem:
Hi! Thanks for the eslint rules. Really helpful.

We found the following example is not caught by the rules I've updated:
```
const myHandler = async (e) => {
  await foo()
  dummy(() => bar(e.currentTarget))
}
```

The reason is, while `await` is in the scope of `myHandler`, `currentTarget` (or `preventDefault()`) is not. To fix it, we have to walk up the scopes and figure out if a parent scope matches the scope where `await` was found.


manuelpuyol:
👋 @kurtextrem thanks for the contribution! Would you mind updating the tests with a new example for this case?

kurtextrem:
> 👋 @kurtextrem thanks for the contribution! Would you mind updating the tests with a new example for this case?

Done!

manuelpuyol (approved):

Issue #539, opened 2024-07-08
Add support for eslint 9


lukasoppermann:
Hey,

this plugin seems to be incompatible with eslint 9 and thus blocks us from upgrading our projects.

It would be great to bring it up to eslint 9 and enable flat config support #513

silverwind:
Was this fixed? At least according to package.json, it's compatible:

https://github.com/github/eslint-plugin-github/blob/abcfc3bdfb4628c75fcbdca1b6fae1fa82cddb1c/package.json#L60

turadg:
Looks like this was a dupe of
- https://github.com/github/eslint-plugin-github/issues/513

which was closed by

- https://github.com/github/eslint-plugin-github/pull/571

WtfJoke:
Yeah it seems to be fixed, because https://github.com/github/eslint-plugin-github/issues/513 is closed and the readme got an entry for flatconfig (eslint9) [here](https://github.com/github/eslint-plugin-github?tab=readme-ov-file#flat-configuration-eslint-configjs)

PR #570, opened 2024-11-06
upgrade the eslint lib to v9 

ejahnGithub:
ref https://github.com/github/eslint-plugin-github/issues/539

upgrade the eslint lib to v9 by following this doc https://eslint.org/docs/latest/use/migrate-to-9.0.0

also updated `context.getScope()` by this https://eslint.org/blog/2023/09/preparing-custom-rules-eslint-v9/#context.getscope() to resolve error

we have both `.eslintrc.js` to support eslint v8 and `eslint.config.mjs` to support eslint v9


gracepark:
I want to preface this by saying I'm still new eslint plugin development (and eslint in general), so there are still some things I'm not 100% sure of. I've come to conclusion that there's 2 separate things here (definitely intertwined with each other still).:

- Support flat config (eslint.config.js)
- Upgrade to eslint v9

This PR upgrades us to eslint v9; however, I think more than upgrading to eslint v9, we want to support flat config. Flat config I believe is available from [8.57.0 and on](https://eslint.org/blog/2024/02/eslint-v8.57.0-released/). So, we don't necessarily _need_ to upgrade to v9 to support flat config. But if we do, even better. 

At the very least, we want to be backwards compatible, so we don't break things for folks who haven't updated to flat config (regardless of version). If we're fine with breaking changes for folks not on flat config, this is a different story, but even so I think we'd still have work to do since our exported configs are not [flat config](https://github.com/github/eslint-plugin-github/blob/main/lib/configs/browser.js).

I think in order for us to be backwards compatible, we'll need to export _both_ legacy and flat configs as mentioned [here](https://eslint.org/docs/latest/extend/plugin-migration-flat-config#backwards-compatibility). In addition, I think we should update our plugin structure.

I found some examples of other plugins that have both configs: 
- https://github.com/amilajack/eslint-plugin-compat/blob/main/src/config/recommended.ts
- https://github.com/import-js/eslint-plugin-import/commit/806e3c2ccc65456a2d8532d575c9f443355bda82#diff-afd32d441615c4b29e00de19367164f9c24769426eea6d2db4bcd759097a2652

I _think_ that may be the reason for some of the issues I ran into. Bc I took [clipboard-copy-element](https://github.com/github/clipboard-copy-element/) which is on v8.42.0 (no flat config). I `npm link`-ed with this PR and I ran into an error, and even if I do install the plugin as mentioned, I have to end up dl-ing every plugin that was referenced in the config file (which I think makes sense):

```
Oops! Something went wrong! :(

ESLint: 8.42.0

ESLint couldn't find the plugin "eslint-plugin-escompat".

(The package "eslint-plugin-escompat" was not found when loaded as a Node module from the directory "/Users/gracepark/github/clipboard-copy-element".)

It's likely that the plugin isn't installed correctly. Try reinstalling by running the following:

    npm install eslint-plugin-escompat@latest --save-dev

The plugin "eslint-plugin-escompat" was referenced from the config file in ".eslintrc.json » plugin:github/browser".

If you still can't figure out the problem, please stop by https://eslint.org/chat/help to chat with the team.
```

And, when I ran it with the current `eslint-plugin-github` package, I did not run into these errors.

This is the section that I think affects us: https://eslint.org/docs/latest/extend/plugin-migration-flat-config#migrating-configs-for-flat-config since we do [export configs](https://github.com/github/eslint-plugin-github/blob/main/lib/configs/browser.js), and those are not flat configs. We'll want both legacy and flat configs.

Let me know if I'm misunderstanding something here.

ejahnGithub:
I’m using `FlatCompat` to handle the transition to `Flat Config`. This allows to integrate legacy `.eslintrc` configurations and plugins without a complete refactor, maintaining compatibility while gradually updating to the Flat Config format.

ref: https://eslint.org/docs/latest/use/configure/migration-guide#using-eslintrc-configs-in-flat-config

PR #578, opened 2024-11-25
Update filenames

gracepark:
The filenames plugin is no longer maintained, and I was running into an issue that `fixupPluginRules` would not fix for us using the compat utility in eslint v9. It needs the `schema`:

https://github.com/github/eslint-plugin-github/pull/578/files#diff-b1acaab48d8385d990537662bf4fa3dd489b8974367f0628147c83977534feb4R13

Since the plugin is no longer being maintained, I decided to bring that particular regex rule into the `github` rules with the fix so that we no longer need to rely on the `filename` plugin.


eslint.config.js
```diff
@@ -23,7 +23,7 @@ module.exports = [
     plugins: {
       eslintPlugin,
       importPlugin,
-      filenames,
+      filenamesPlugin: fixupPluginRules(filenames),
```
arelia, line 26: does it matter that it's not a string like `'i18n-text'` on the next line?

arelia (approved):
added one comment i'm not sure about but otherwise looks good

eslint.config.js
gracepark: That should be okay - what I ended up doing was bringing in the filenames match regex rule into github. The filenames plugin is no longer supported and the compat utility was not working for our rule configuration.

docs/rules/filenames-match-regex.md
```diff
@@ -0,0 +1,35 @@
+# Ensure filenames match a regex naming convention (`github/filenames-match-regex`)
+
+💼 This rule is enabled in the ✅ `recommended` config.
+
+<!-- end auto-generated rule header -->
+
+## Rule Details
+
+TRule to ensure that filenames match a convention, with a default of camelCase.
```
arelia, line 9:
```suggestion
Rule to ensure that filenames match a convention, with a default of camelCase.
```

arelia (commented):
looks good, but I'm not familiar with how GitHub does open source so I wonder if it matters that the filenames plugin didn't have a LICENSE.md  and is there anything else we should do when bringing it over. It might be worth pinging the team

arelia (approved):

PR #597, opened 2025-01-07
fix: upgrade eslint-plugin-jsx-a11y

hyldmo:
While it seems most of the configs have been updated to work with eslint 9, the dependency https://github.com/github/eslint-plugin-github/blob/1b65401f6ede80ecdc0da5398043081cd6e2b8f3/package.json#L47
doesn't have support for flat configs until [v6.10.0](https://github.com/jsx-eslint/eslint-plugin-jsx-a11y/releases/tag/v6.10.0).

This means that  it crashes here: https://github.com/github/eslint-plugin-github/blob/1b65401f6ede80ecdc0da5398043081cd6e2b8f3/lib/configs/flat/react.js#L6

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -22,7 +22,7 @@
         "eslint-plugin-filenames": "^1.3.2",
         "eslint-plugin-i18n-text": "^1.0.1",
         "eslint-plugin-import": "^2.25.2",
-        "eslint-plugin-jsx-a11y": "^6.7.1",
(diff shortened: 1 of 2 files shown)


hyldmo:
A small side note I think the same versioning issue applies to` eslint-plugin-import`. Note: I'm using yarn, which afaik has a bit different rules on how to resolve versions

manuelpuyol (approved):

PR #605, opened 2025-01-30
Update names back to original for eslint-comments and no-only-tests

gracepark:
Updating names back to original for the plugin.


lib/configs/flat/recommended.js
```diff
@@ -17,21 +17,21 @@ module.exports = {
   },
   plugins: {
     prettierPlugin,
-    eslintComments,
+    'eslint-comments': eslintComments,
```
jibrang, line 20: Just curious. Why is it necessary to revert to the original ones?

gracepark: This way we can use the same names in dotcom. Or else we'd have to update the names in the monolith to `eslintComments`.

jibrang: Thanks for clarifying!

PR #608, opened 2025-02-03
Update a11y rule config to not flag `role` being set for `ul`

khiga8:
This PR updates the config for the [no-redundant-roles](https://github.com/jsx-eslint/eslint-plugin-jsx-a11y/blob/main/docs/rules/no-redundant-roles.md) lint rule.

In webkit, setting `list-style-type: none` results in semantics being removed, so explicit roles are necessary.

### DIFF
--- a/lib/configs/react.js
+++ b/lib/configs/react.js
@@ -36,7 +36,7 @@ module.exports = {
         nav: ['navigation'], // default in eslint-plugin-jsx-a11y
         tbody: ['rowgroup'],
         thead: ['rowgroup'],
-        ul: ['list'], // In webkit, setting list-style-type: none results in semantics being removed. Need for explicit role.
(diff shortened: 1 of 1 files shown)


manuelpuyol (approved):

lib/configs/react.js
```diff
@@ -36,6 +36,7 @@ module.exports = {
         nav: ['navigation'], // default in eslint-plugin-jsx-a11y
         tbody: ['rowgroup'],
         thead: ['rowgroup'],
+        ul: ['list'], // In webkit, setting list-style-type: none results in semantics being removed. Need for explicit role.
```
JoyceZhu, line 39:
```suggestion
        ul: ['list'], // In webkit, setting list-style-type: none results in semantics being removed. Need explicit role.
```

JoyceZhu (approved):

PR #609, opened 2025-02-04
Update prettier plugin name

gracepark:
This PR changes the prettier plugin name to `prettier` in the `lib/configs/flat/recommended.js` file so we can keep the same name as before.


bteng22 (approved):
LGTM!

PR #621, opened 2025-03-19
chore: update rule meta to satisfiy lint rules

43081j:
Updates a few things in rule meta so the linter is happy:

- Some descriptions didn't follow the naming convention (`enforce`, `require`, etc)
- Some rules had no `type`, so these were set to `problem`
- Some rules had no `recommended`, so these were set to if they are in the recommended config or not
- Message IDs have been used instead of messages

these didn't happen in CI but did error locally for whatever reason

edit:

so it turns out the eslint-plugin-eslint-plugin rules haven't been running against the rule files

for whatever reason, `ast.sourceType` ends up as `"module"` inside the plugin, so it tries to look for ESM exports and doesn't find any. so the rules never run


keithamus (approved):
This looks like a great cleanup, thanks so much @43081j!

PR #622, opened 2025-03-19
feat: migrate to ESM only

43081j:
All LTS versions of node will soon support `require(esm)` as `18.x` will fall out of long term support imminently

ESLint also already loads plugins via a dynamic import, so if we make this change it means the only people affected are those writing flat configs as CommonJS _in Node <20_

Depends on #621 since this makes the lint rules apply again (they currently don't in main)


keithamus (approved):
This LGTM.

PR #686, opened 2026-02-25
fix: add ESLint v10 to peerDependencies

roli-lpci:
## Summary

Widens `eslint` peerDependencies from `^8 || ^9` to `^8 || ^9 || ^10`.

All rules already use the modern `context.sourceCode` / `sourceCode.getScope()` APIs with fallbacks for older versions, so no runtime changes are needed. Audited all 5 rule files that reference deprecated context APIs — all already use `??` / ternary fallback patterns.

Closes #680

## Test plan

- [x] All 67 existing tests pass
- [x] ESLint linting passes
- [x] Audited all rule files using deprecated context APIs — all already use fallback patterns

PR #702, opened 2026-05-12
fix: Updates to add eslint10 to peer dependencies

CharlieM312:
Added eslint-10 to peer dependencies and updated the relevant rules based on the feedback I saw in #686 
Let me know if this needs any additional changes to it, I believe the rules may need updating.
Also fixed an issue in get-attribute.js that was causing the eslint-check script to fail.


dgreif:
Closing in favor of https://github.com/github/eslint-plugin-github/pull/713

PR #705, opened 2026-06-03
Update Node.js version and checkout action versions

gracepark:
Part of: https://github.com/github/web-systems/issues/4309#issuecomment-4608163612

OIDC migration


dgreif:
@gracepark sorry for the double work here, I just merged https://github.com/github/eslint-plugin-github/pull/704 which tackled these same bits, plus some other supply chain improvements

